Issue #5, opened 2019-02-01
Wrap telepresence in CLI command


bartoszmajsak:
As a first cut, following command:
```
$ ike serve --deployment name --port 8000 --run 'python3 server.py'` 
```

should result in invoking telepresence as part of it

```
telepresence --swap-deployment name --expose 8000 --method inject-tcp --run python3 server.py;'
```

PR #8, opened 2019-02-01
feat(#5): wraps telepresence in ike develop cmd

bartoszmajsak:
```
Usage:
  ike develop [flags]

Flags:
  -d, --deployment string   name of the deployment or deployment config
  -h, --help                help for develop
  -m, --method string       telepresence proxying mode - see https://www.telepresence.io/reference/methods (default "inject-tcp")
  -p, --port int            port to be exposed (default 8000)
  -r, --run string          command to run your application
```

### DIFF
--- a/cmd/ike/cmd/develop.go
+++ b/cmd/ike/cmd/develop.go
@@ -39,7 +39,9 @@ var developCmd = &cobra.Command{
 	Long:  `TODO`,
 	Run: func(cmd *cobra.Command, args []string) {
 
-		checkIfTelepresenceExists()
+		if !telepresenceExists() {
+			os.Exit(1)
+		}
 
 		var tp = exec.Command(telepresenceBin, parseArguments()...)
 
@@ -94,14 +96,15 @@ func parseArguments() []string {
 		"--run"}, runArgs...)
 }
 
-func checkIfTelepresenceExists() {
+func telepresenceExists() bool {
 	path, err := exec.LookPath(telepresenceBin)
 	if err != nil {
 		log.Error(err, fmt.Sprintf("Couldn't find '%s' installed in your system.\n"+
 			"Head over to https://www.telepresence.io/reference/install for installation instructions.\n", telepresenceBin))
-		os.Exit(1)
+		return false
 	} else {
 		log.Info(fmt.Sprintf("Found '%s' executable in '%s'.", telepresenceBin, path))
 		log.Info(fmt.Sprintf("See '%s.log' for more details about its execution.", telepresenceBin))
+		return true
 	}
 }
--- a/cmd/ike/cmd/develop.go
+++ b/cmd/ike/cmd/develop.go
@@ -21,6 +21,10 @@ var (
 )
 
 func init() {
+	// -- config (bound with viper?)
+	// -- build command
+	// --watch to rebuild ---> see skaffold samples for node and python
+	// -- testing ?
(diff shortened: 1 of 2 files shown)


cmd/ike/main.go
```diff
@@ -10,7 +10,7 @@ func main() {
 	// implementing the logr.Logger interface. This logger will
 	// be propagated through the whole operator, generating
 	// uniform and structured logs.
-	logf.SetLogger(logf.ZapLogger(false))
+	logf.SetLogger(logf.ZapLogger(true))
```
aslakknutsen: Need a better way to handle this as we want the Service to deploy using 'false'

cmd/ike/cmd/develop.go
```diff
@@ -0,0 +1,106 @@
...
+
+const telepresenceBin = "telepresence"
+
+var developCmd = &cobra.Command{
+	Use:   "develop",
+	Short: "starts the development flow",
+	Long:  `TODO`,
+	Run: func(cmd *cobra.Command, args []string) {
+
+		checkIfTelepresenceExists()
```
aslakknutsen: Shouldn't this 'stop' the further execution?

cmd/ike/cmd/develop.go
```diff
@@ -0,0 +1,106 @@
...
+		log.V(9).Info("Failed to copy either of stdout or stderr")
+	}
+
+	wg.Wait()
+}
+
+func parseArguments() []string {
+	runArgs := strings.Split(runnable, " ")
+	return append([]string{
+		"--swap-deployment", deploymentName,
```
aslakknutsen: Should we write down an issue on not using swap from the client, but having the service prepare the 'correct' image?

cmd/ike/main.go
bartoszmajsak: You want that to be fixed as part of this PR? Maybe we need two separated things - CLI and that thing which runs on the server? That's probably a decision which needs to be discussed in a bit more detail. For now I'm fine with `json` output - can revert it to the previous state and make an issue for the potential split. WDYT?

cmd/ike/cmd/develop.go
bartoszmajsak: Right, with `logrus` it did without explicit `exit` call. Added in 9f08970

cmd/ike/cmd/develop.go
bartoszmajsak: I can take care of it today.

cmd/ike/main.go
aslakknutsen: sounds good.

cmd/ike/cmd/develop.go
```diff
@@ -0,0 +1,107 @@
...
+		"--method", method,
+		"--run"}, runArgs...)
+}
+
+func checkIfTelepresenceExists() {
+	path, err := exec.LookPath(telepresenceBin)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Couldn't find '%s' installed in your system.\n"+
+			"Head over to https://www.telepresence.io/reference/install for installation instructions.\n", telepresenceBin))
+		os.Exit(1)
```
aslakknutsen: ... technically I would return an error and handle the exit in the caller.. or rename the function to explicitly 'warn' about the exit happening.

bartoszmajsak: How about meeting mid-way - ded1887?

cmd/ike/cmd/develop.go
bartoszmajsak: Created epic #10

PR #21, opened 2019-02-08
feat(config): introduces configuration complementary to flags

### LINKED ISSUE #12: Provide ability to specify command configuration from other sources than their flags
Instead of relying on CLI flags we should give the user the ability to specify common / infrequently changed configuration items through other means, such as environment variable (prefixed with `IKE_`) and yaml file.

So instead of e.g. invoking `ike develop --deployment NAME --method inject-tcp` developer can set `IKE_DEVELOP_DEPLOYMENT=NAME` and have config file (`.ike.config` sitting in the folder from which the command is invoked) with content:

```yaml
develop:
  method: inject-tcp
```

instead.

Following order of precedence should be preserved when calling a command (where each item takes precedence over the item below, e.g. flag will always be the most important one):

* flag
* environment variable
* config file entry
* default (if specified)

bartoszmajsak:
### Changes in this PR

  * external config sources based on [viper](https://github.com/spf13/viper) (in order of importance)
    * flags
    * env variables
    * config file
  * test setup based on [ginkgo](https://github.com/spf13/viper)
  * refactored commands to not be singletons
  * introduces `config` packages which wraps viper
    * centralized settings
    * keeping config and flags in sync (so we still can use required flag validation)

### Rationale

Common settings, such as `run` and `build` recipes, could be shared in the repo as a config file rather than being used constantly from the CLI

Fixes #12

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -73,7 +73,8 @@ required = [
   name = "github.com/onsi/ginkgo"
   version = "1.7.0"
 
-# Apply workaround from https://github.com/golang/dep/issues/1799
+# Workaround for https://github.com/golang/dep/issues/1799
+# Otherwise ginkgo/gomega won't be pulled in
 [[override]]
   name = "gopkg.in/fsnotify.v1"
   source = "https://github.com/fsnotify/fsnotify.git"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -9,6 +9,14 @@ required = [
   "k8s.io/gengo/args",
 ]
 
+[prune]
+  go-tests = true
+  non-go = true
+
+  [[prune.project]]
+    name = "k8s.io/code-generator"
+    non-go = false
+
 [[override]]
   name = "k8s.io/code-generator"
   # revision for tag "kubernetes-1.13.1"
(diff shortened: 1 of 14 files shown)


Gopkg.toml
```diff
@@ -57,10 +65,15 @@ required = [
...
+  version = "1.4.3"
 
-  [[prune.project]]
-    name = "k8s.io/code-generator"
-    non-go = false
+[[constraint]]
+  name = "github.com/onsi/ginkgo"
+  version = "1.7.0"
+
+# Apply workaround from https://github.com/golang/dep/issues/1799
```
aslakknutsen: not part of PR ?

bartoszmajsak: Maybe comment is misleading. This issue is still open and without this constraint, it fails fetching ginkgo/gomega. I will reword the comment to avoid confusion.

bartoszmajsak:
Trying to figure out why tests are failing on Travis...

Gopkg.toml
bartoszmajsak: Addressed in b83b9b3.

PR #25, opened 2019-02-12
chore(errors): cmds to return errors instead of exiting

bartoszmajsak:
Instead of using `Run` and regular Cobra hooks, let's start using `RunE` and alike to return errors instead of calling `os.Exit` and let Cobra handle that instead.

This gives us the immediate benefit of better testability - as there's no `os.Exit` in the code under test, there is no risk in abruptly finishing test suite for example.

#### Open Question

As-is Cobra will print the usage of the cmd in case of failure. This can be easily disabled by setting `SilenceUsage` flag to `false`. Shall we consider that change too? @aslakknutsen

### DIFF
--- a/cmd/ike/cmd/develop.go
+++ b/cmd/ike/cmd/develop.go
@@ -33,14 +33,16 @@ func NewDevelopCmd() *cobra.Command {
 
 			return nil
 		},
-		Run: func(cmd *cobra.Command, args []string) { //nolint[:unparam]
+		RunE: func(cmd *cobra.Command, args []string) error { //nolint[:unparam]
 			var tp = exec.Command(telepresenceBin, parseArguments(cmd)...)
-			redirectStreams(tp)
-			err := tp.Wait()
-			if err != nil {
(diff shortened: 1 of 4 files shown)


aslakknutsen:
Yeah, it's a little annoying getting cli flag help when it's a server error.

Any chance of not silence usage based on type of error, e.g. flag error is ok, 'run' error is not ?

bartoszmajsak:
I will have a look

bartoszmajsak:
So here's how it works. First the flags are parsed, if an error occurs, and `SilenceErrors` is set to `true`, it's printed and `--help` is suggested for a given command.

Then the actual execution happens. Here's the flow:

* if `--help` is requested - it shows it regardless
* if `SilenceErrors` is set to `true` the error is shown (that's "server" error)
* if `SilenceUsage` is set it will prompt `usage` message

So I think we are good with setting `SilenceUsage` to false. WDYT?

aslakknutsen:
@bartoszmajsak yeah, sounds good.

bartoszmajsak:
Done in d5cdf89

PR #28, opened 2019-02-12
test(telepresence): tests proper parameter delegation

bartoszmajsak:
Adds tests verifying if we pass arguments to `telepresence` correctly.

This is achieved by swapping actual `telepresence` binary with test double - simple executable printing out all the arguments which were passed. That's achieved by Ginko's `gexec.Build` invoked before suite which builds a simple program included in this PR (`test/echo/telepresence/main.go`) and makes it `telepresence` command available in the `$PATH`.

In addition structure of Specs has been slightly changed.

### DIFF
--- a/cmd/ike/cmd/develop_test.go
+++ b/cmd/ike/cmd/develop_test.go
@@ -4,6 +4,9 @@ import (
 	"os"
 	"path"
 
+	"github.com/onsi/gomega/gexec"
+	"github.com/spf13/afero"
+
 	. "github.com/aslakknutsen/istio-workspace/cmd/ike/cmd"
 
 	. "github.com/aslakknutsen/istio-workspace/test"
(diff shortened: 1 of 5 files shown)


cmd/ike/cmd/develop_test.go
```diff
@@ -26,23 +28,23 @@ var _ = Describe("Usage of ike develop command", func() {
...
-		telepresenceBin := TmpFile(GinkgoT(), "telepresence", "echo 'telepresence'")
-		_ = os.Setenv("PATH", path.Dir(telepresenceBin.Name()))
+		telepresenceBin, err := gexec.Build("github.com/aslakknutsen/istio-workspace/test/echo/telepresence")
+		Expect(err).ToNot(HaveOccurred())
+		_ = os.Setenv("PATH", path.Dir(telepresenceBin))
 	})
 
 	AfterSuite(func() {
 		_ = os.Setenv("PATH", originalPath)
-		CleanUp(GinkgoT())
```
bartoszmajsak: This was moved to `AfterEach` in config spec.

PR #34, opened 2019-02-19
feat(develop): introduces watch functionality

### LINKED ISSUE #22: Watching file changes to re-run the dev mode process
When changes are made on the files in the specified locations we should rebuild the project if defined how and re-run telepresence session.

This should be enabled if `ike develop` has `--watch` flag provided (or is set in the cfg file). This should be a list of directory patterns to observe and re-run the build and telepresence process when a change occured.

By default both `build` and `run` steps are executed when calling `ike develop`, unless `--run-only` flag is specified.

bartoszmajsak:
#### Why

To seamlessly restart `telepresence` when changes occur.

#### What

* `ike develop --watch [directories] --watch-exclude [glob patterns]`
* `ike develop --build [command]`

#### What's not in (yet)

* `.gitignore` support - if present in the directory should be added to exclusions

#### Testing

There's one special group of tests which I want to elaborate on a litte. 

When a file on observed path changes, watch does following:

  * re-runs `build` command (unless `--no-build` specified)  (in our tests we use `mvn`)
  * stops `telepresence` process (if it's running)
  * starts `telepresence` process
  
For testing purposes, I've introduced a simple [`echo` program](https://github.com/aslakknutsen/istio-workspace/blob/54908444413cf03da1751d5b95fe1e338cf5aa71/test/echo/main.go) which prints out passed commands. This is later used in tests to verify if the particular command was executed. For example when we change file under watch, we expect that `mvn clean install` was invoked twice. We use this binary in three flavors:

   * as stub for `mvn`
   * as stub for `telepresence`
   * as stub with delay for `telepresence` - to simulate the running process. This simply sets process to sleep for defined duration before exiting

Those binaries as compiled `BeforeSuite` and used across test cases through `$PATH` substitution.

https://github.com/aslakknutsen/istio-workspace/blob/54908444413cf03da1751d5b95fe1e338cf5aa71/cmd/ike/cmd/develop_test.go#L36-L41

In case of watch test cases, I couldn't find a better approach than:

  * start "delayed telepresence" in a goroutine
    * on completion sends output to test channel used in the test
  * sleeps for few milliseconds (but shorter than "delayed telepresence" runs
  * makes changes in the watched file
  * waits until channel receives output and inspect what was printed

See concrete example:

https://github.com/aslakknutsen/istio-workspace/blob/54908444413cf03da1751d5b95fe1e338cf5aa71/cmd/ike/cmd/develop_test.go#L266-L292

Closes #22


cmd/ike/cmd/develop.go
```diff
@@ -2,53 +2,106 @@ package cmd
...
+
+			if cmd.Flag("watch").Changed {
+				slice, e := cmd.Flags().GetStringSlice("watch")
+				if e != nil {
+					return e
+				}
+
+				w, err := watch.NewWatch().
+					WithHandler(func(event fsnotify.Event, done chan<- struct{}) error {
+						_, _ = cmd.OutOrStdout().Write([]byte(event.Name + " changed. Restarting process.\n"))
```
bartoszmajsak: This is temporary workaround to be able to inspect the output in the buffer used in tests.

I will work on the proper solution next.

bartoszmajsak:
#### Tasks

- [ ] batch `fsnotify` events to avoid too many restarts
- [ ] fix logging / output inspection for testing https://github.com/aslakknutsen/istio-workspace/pull/34#discussion_r258048595
- [ ] rework watch to not restart the whole tp process, but be the command executed by `--run` from telepresence instead
- [ ] explain how to pass multi-value flags (e.g. `--run`)
- [ ] **NICE TO HAVE** smoke test agaisnt real cluster and real TP

cmd/ike/cmd/develop.go
```diff
@@ -107,6 +109,7 @@ func NewDevelopCmd() *cobra.Command {
 	developCmd.Flags().Bool("no-build", false, "always skips build")
 	developCmd.Flags().StringSliceP("watch", "w", []string{currentDir()}, "list of directories to watch")
 	developCmd.Flags().StringSlice("watch-exclude", excludeTpLog, "list of patterns to exclude (defaults to telepresence.log which is always excluded)")
+	developCmd.Flags().Int64("watch-interval", 500, "watch interval (in ms)")
```
aslakknutsen: Don't think I would have exposed this. Does a user really care?

cmd/ike/cmd/develop.go
```diff
@@ -107,6 +109,7 @@ func NewDevelopCmd() *cobra.Command {
 	developCmd.Flags().Bool("no-build", false, "always skips build")
```
aslakknutsen: why a -build and a -no-build? Why isn't "no-build" just not having a -build?

cmd/ike/cmd/develop.go
```diff
@@ -107,6 +109,7 @@ func NewDevelopCmd() *cobra.Command {
...
 	developCmd.Flags().StringSliceP("watch", "w", []string{currentDir()}, "list of directories to watch")
 	developCmd.Flags().StringSlice("watch-exclude", excludeTpLog, "list of patterns to exclude (defaults to telepresence.log which is always excluded)")
```
aslakknutsen: how about *.log? But I guess default should be .gitignore?

cmd/ike/cmd/develop_test.go
```diff
@@ -268,6 +268,8 @@ var _ = Describe("Usage of ike develop command", func() {
 					"--build", "mvn clean install",
 					"--port", "4321",
```
aslakknutsen: We can see if we can automap the port based on DC config or Run config

cmd/ike/cmd/develop.go
bartoszmajsak: Yeah, I was more thinking about handling `.gitignore` and assume it's there. But we can always have set of defaults regardless. That makes sense.

cmd/ike/cmd/develop.go
bartoszmajsak: b/c you might have it in the config - still I can try to resolve that case but thought that would be more explicit.

cmd/ike/cmd/develop.go
bartoszmajsak: Right, maybe it's only useful for tests.

cmd/ike/cmd/develop_test.go
bartoszmajsak: Could you open issue for it?

cmd/ike/cmd/develop.go
bartoszmajsak: Made it hidden 7626fe9

cmd/ike/cmd/develop.go
bartoszmajsak: Made it hidden 7626fe9

cmd/ike/cmd/develop.go
aslakknutsen: ah, ic...

cmd/ike/cmd/develop_test.go
aslakknutsen: https://github.com/aslakknutsen/istio-workspace/issues/35

PR #44, opened 2019-02-25
feat(develop): allow to define local:remote port to run multiple serv…

aslakknutsen:
…ices on localhost

### DIFF
--- a/cmd/ike/cmd/develop.go
+++ b/cmd/ike/cmd/develop.go
@@ -54,7 +54,7 @@ func NewDevelopCmd() *cobra.Command {
 	}
 
 	developCmd.Flags().StringP("deployment", "d", "", "name of the deployment or deployment config")
-	developCmd.Flags().IntP("port", "p", 8000, "port to be exposed")
+	developCmd.Flags().StringP("port", "p", "8000", "port to be exposed in format local[:remote]")
 	developCmd.Flags().StringP(runFlagName, "r", "", "command to run your application")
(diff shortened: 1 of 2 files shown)


cmd/ike/cmd/develop.go
```diff
@@ -54,7 +54,7 @@ func NewDevelopCmd() *cobra.Command {
 	}
 
 	developCmd.Flags().StringP("deployment", "d", "", "name of the deployment or deployment config")
-	developCmd.Flags().IntP("port", "p", 8000, "port to be exposed")
+	developCmd.Flags().StringP("port", "p", "8000", "port to be exposed in format local[:remote]")
```
bartoszmajsak: Might make sense to adjust one of the tests here:

https://github.com/aslakknutsen/istio-workspace/blob/master/cmd/ike/cmd/develop_test.go#L157-L189

This also opens up a question if we want to support multiple ports? If so - I will open an issue for that. See https://github.com/telepresenceio/telepresence/issues/280

aslakknutsen: Yea, multi port makes sense

aslakknutsen: addressed https://github.com/aslakknutsen/istio-workspace/pull/44/commits/d6a8f7c65c249e40a8b3f07a28e4a8e6853ccec9

bartoszmajsak: #45

PR #48, opened 2019-02-27
test(e2e): introduces basic e2e smoke tests

### LINKED ISSUE #41: Smoke test against real cluster and real telepresence

bartoszmajsak:
Introduces smoke e2e tests against maistra (using `istiooc` to start up the cluster).

##### Covered scenarios

  * simple `ike develop` 
  * `ike develop --watch`

Fixes #41

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,7 @@ jobs:
     machine:
       image: circleci/classic:201808-01
     steps:
+      - run: sudo rm -rf /usr/local/go
       - run: sudo circleci-install golang 1.12
       - checkout
       - run: go env
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,7 @@ workflows:
       - build
   circleci_e2e_tests:
     jobs:
+      - build
       - e2e_tests:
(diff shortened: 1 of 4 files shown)


bartoszmajsak:
Putting it as draft b/c CI stage is needed too.

bartoszmajsak:
Sorry for the humongous amount of commits - I was "testing" / fixing circle-ci setup. Got a bit of corner cases to solve on the way. Cool thing I learned today are "orbs" - reusable commands. I created one for this project - https://github.com/bartoszmajsak/circleci-orbs (we can move it to a new home at some point).

We have `circleci` jobs working right now - https://circleci.com/gh/bartoszmajsak/istio-workspace/176

What we have to do is to enable it for that repo. Only you can do that though, unless we move it to `Maistra` org.

#### Open points

* we have free 1000 minutes of build time per month per org - e2e build takes now 9 minutes. This gives us ~ 100-110 builds per month.  One solution would be to somehow trigger e2e only based on commit hashtag or sth - I will think about it
* there are few things we can perhaps optimize in circleci - sharing between workspaces, caching etc

Issue #56, opened 2019-03-04
Update e2e to use bookinfo and use ike operator


aslakknutsen:
Currently, e2e rely on --offline to avoid Session creation

Should be updated to do:

make deploy-operator
make deploy-sample-bookinfo (missing. services, gateway, virtualservices, destinationrules)
ike develop --deployment details-v1?

PR #65, opened 2019-03-05
chore(rand_name): moves random name generator to its own pkg

bartoszmajsak:
and adds unit tests for it (with custom gomega matchers)

(ideally these should be property-based tests, but I will work on that when done with #56)


pkg/naming/name_generator_test.go
```diff
@@ -0,0 +1,51 @@
...
+	})
+
+	It("should always generate letter when single character name is requested", func() {
+		name := naming.RandName(1)
+		Expect(name).To(OnlyContain("abcdefghijklmnopqrstuvwxyz"))
+	})
+
+	It("should generate name with first character being a letter", func() {
+		name := naming.RandName(32)
+		Expect(name).To(StartWithLetter())
```
aslakknutsen: these are a bit random if they fail or not are they not?

bartoszmajsak: If there's a bug - yes. But there is not :)

As I said in the PR description - ideally it should be done using generators, but I would rather first ship that and then use generators to improve tests.

aslakknutsen: Why not just remove '-' from the options. It's not like it's required :)

bartoszmajsak: Still, cannot be a number

bartoszmajsak: We could simplify though and remove those tests, true. You favor this over what we have now?

aslakknutsen: we don't need to use numbers either.. :)

bartoszmajsak: ACK

bartoszmajsak: Fixed with 862840b. I just wanted it to look fancy :P

PR #77, opened 2019-04-02
chore(deps): updates operator-sdk to 0.6.0

bartoszmajsak:


bartoszmajsak:
Circle has some infra issues again. Failing while trying to install telepresence.

```
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
Exited with code 100
```

PR #87, opened 2019-04-04
feat(#75): simplifies e2e boostrap

### LINKED ISSUE #75: Streamline cluster setup process
It turned out we have to make quite a bit of cluster mgmt commands before we are able to execute e2e tests on circleci. At this point what was supposed to be simple setup/teardown part of e2e-tests in ginkgo code turned out to grow a bit too much.

We should move most of the cluster setup into make file instead, so that we can conveniently use it both for CI and while working locally with the code.

bartoszmajsak:
This PR brings following improvements in e2e tests setup:

* starts cluster only when needed (if not running)
  * allows to skip shutdown if `SKIP_CLUSTER_SHUTDOWN` is set
* reworks cluster deployments to be in `Makefile` (including bookinfo app)
* deploys istio operator once per test suite
* logs more detailed checks to quickly see if ci build is not hanging
* prints cmds as plain string

Fixes #75


e2e/e2e_suite_test.go
```diff
@@ -42,19 +43,28 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 			fmt.Printf("\nExposing Docker Registry\n")
 			<-cmd.Execute("oc", "expose", "service", "docker-registry", "-n", "default").Done()
 
-			// create a 'real user' we can use to push to the DockerRegsitry
+			// create a 'real user' we can use to push to the DockerRegistry
 			fmt.Printf("\nAdd admin user\n")
 			<-cmd.Execute("oc", "create", "user", "admin").Done()
 			<-cmd.Execute("oc", "adm", "policy", "add-cluster-role-to-user", "cluster-admin", "admin").Done()
```
aslakknutsen: Any reason to keep these in code?

bartoszmajsak: I thought it's specific to our tests, so whoever is using `Makefile` can use any user / privileges they want

PR #89, opened 2019-04-08
feat(#76): brings go client to interact with Sessions

### LINKED ISSUE #76: Replace direct oc cli calls with the golang client
Replace https://github.com/aslakknutsen/istio-workspace/blob/c6e328d39196698bb67fcd25a831a82d52f80ab1/pkg/istio/cli_helper.go with go client.

###### Currently used

* `oc apply` to create session
* `oc get session`
* `oc delete session`

bartoszmajsak:
Fixes #76 

The reason why the PR is marked as DRAFT:

- [x] support for flags when no `.kube/config` is present

Currently when loading cluster config we look for `.kube/config` file unless `KUBECONFIG` env variable is set. We should add extra flags to handle a scenario when we start from scratch without those files being present. 

**Q:** Would `--kubeconfig` flag pointing at a file be sufficient here or do we want to break it down to all required params? Or maybe for now `KUBECONFIG` (documented) would be sufficient to move on?

- [x] basic unit tests for session handling logic

working on using generated fake-client for that


Makefile
```diff
@@ -54,21 +54,27 @@ lint: deps ## Concurrently runs a whole bunch of static analysis tools
...
+GROUP_VERSIONS:="istio:v1alpha1"
 .PHONY: codegen
 codegen: install-operator-sdk ## Generates operator-sdk code
 	$(call header,"Generates operator-sdk code")
 	$(CUR_DIR)/bin/operator-sdk generate k8s
+	$(call header,"Generates clientset code")
+	GOPATH=$(shell echo ${GOPATH} | rev | cut -d':' -f 1 | rev) ./vendor/k8s.io/code-generator/generate-groups.sh client \
+		$(PACKAGE_NAME)/pkg/client \
+		$(PACKAGE_NAME)/pkg/apis \
+		$(GROUP_VERSIONS)
```
bartoszmajsak: @aslakknutsen Now we have both `operator-sdk` CLI which provides "deepcopy" generator and this which can also provide the same. Do we want to keep that as is or unify to only use `./vendor/k8s.io/code-generator/generate-groups.sh`

cmd/ike/session/session.go
```diff
@@ -26,13 +22,23 @@ func Offline(cmd *cobra.Command) (func(), error) {
...
+var sessionClient *client
+
 // CreateOrJoinHandler provides the option to either create a new session if non exist or join an existing.
 // Rely on the following flags:
 //  * deployment - the name of the target deployment and will update the flag with the new deployment name
 //  * session - the name of the session
 func CreateOrJoinHandler(cmd *cobra.Command) (func(), error) {
 	sessionName := getSessionName(cmd)
 	deploymentName, _ := cmd.Flags().GetString("deployment")
+	if sessionClient == nil {
```
bartoszmajsak: Not particularly happy with this "lazy loading" but couldn't come up with a better approach (besides wrapping it up in a `func` and use that instead of `var` reference in all the `func`s here). Any suggestions on how we could make it better?

cmd/ike/session/session_client.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (c *client) Delete(session *istiov1alpha1.Session) error {
+	if err := c.client.IstioV1alpha1().Sessions(c.namespace).Delete(session.Name, &metav1.DeleteOptions{}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *client) getSession(sessionName string) (*istiov1alpha1.Session, error) {
```
bartoszmajsak: Maybe we export it and simply call `Get` ? WDYT?

cmd/ike/session/session.go
aslakknutsen: Could move the instance to where it's created like with http.DefaultClient type of thing.

Move sessionClient to session_client and lazy it in there. Separate between DefaultClient() and NewClient() or similar. Basically the same, but less scattered.

Makefile
aslakknutsen: hmm... Any advantage of using the Generated client vs just a client.Client? 'Type safe' Get?

bartoszmajsak: "type safe" get/create/delete :)

plus dogfooding potentially used stuff by other "clients"

cmd/ike/session/session.go
bartoszmajsak: Done 711872a

cmd/ike/session/session_client.go
bartoszmajsak: Done fdb32e3

aslakknutsen:
KUBECONFIG would be fine for now I think.

PR #92, opened 2019-04-11
chore(gofmt): fixes missed formatting

bartoszmajsak:
I can work on a step in CI which would automate that


bartoszmajsak:
Opened #93

PR #99, opened 2019-04-16
feat(route): Add API to define how to route traffic

### LINKED ISSUE #60: Create a API for strategy choice around route matching
* --header "end-user:jason" 
* ?

- [ ] document api
- [ ] document user guide

aslakknutsen:
fixes #60

- [x] Service
- [x] CLI


cmd/ike/cmd/develop_test.go
```diff
@@ -87,6 +87,12 @@ var _ = Describe("Usage of ike develop command", func() {
 				Expect(developCmd.Flag("method").Value.String()).To(Equal("inject-tcp"))
 			})
 
+			It("should be able to provide the traffic route parameter", func() {
+				_, err := ValidateArgumentsOf(developCmd).Passing("--deployment", "rating-service", "--run", "java", "--route", "header:name=value")
+
+				Expect(err).NotTo(HaveOccurred())
```
bartoszmajsak:
```suggestion
				Expect(err).NotTo(HaveOccurred())
				Expect(developCmd.Flag("route").Value.String()).To(Equal("header:name=value"))
```

cmd/ike/cmd/develop.go
```diff
@@ -77,6 +77,7 @@ func NewDevelopCmd() *cobra.Command {
 	}
 	developCmd.Flags().StringP("method", "m", "inject-tcp", "telepresence proxying mode - see https://www.telepresence.io/reference/methods")
 	developCmd.Flags().StringP("session", "s", "", "create or join an existing session")
+	developCmd.Flags().StringP("route", "", "", "specify traffic route options in the format of type:name=value")
```
bartoszmajsak: Is there any default value if not specified? If so maybe worth mentioning it in the help text.

cmd/ike/session/session_test.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"github.com/aslakknutsen/istio-workspace/cmd/ike/session"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Session operations", func() {
+
+	Context("route parsing", func() {
```
bartoszmajsak: This actually would be awesome to test using something like [Gopter](https://github.com/leanovate/gopter) (property based testing). I will take a look in my spare time.

cmd/ike/session/session_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+		It("should return nil with no error on empty string", func() {
+			r, err := session.ParseRoute("")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(r).To(BeNil())
+		})
+
+		It("should error on wrong type format", func() {
+			_, err := session.ParseRoute("a=b")
+			Expect(err).To(HaveOccurred())
```
bartoszmajsak: How about checking on the message too? As we don't have error types it might at least ensure we actually cover the expected error path.

I can take care of it if you agree.

pkg/controller/session/model_convert_test.go
```diff
@@ -153,4 +155,52 @@ var _ = Describe("Basic model conversion", func() {
 		})
 
 	})
+	Context("route to route", func() {
+		var (
+			route model.Route
+		)
+		JustBeforeEach(func() {
+			route = session.RouteToRoute(&sess)
```
bartoszmajsak: I think this should belong to the test itself. From my point of view before/after are for test fixtures / setup / cleanup, but not for the interaction with the actual component under test.

cmd/ike/session/session_test.go
aslakknutsen: Good idea.

cmd/ike/session/session_test.go
aslakknutsen: Sure, could do that.

pkg/controller/session/model_convert_test.go
aslakknutsen: Just as a note; that would include all the model tests.

cmd/ike/cmd/develop.go
aslakknutsen: It defaults on the Server SIde: https://github.com/aslakknutsen/istio-workspace/pull/99/files#diff-41947331b7e282077cf9be05075567fdR66

cmd/ike/session/session_test.go
bartoszmajsak: #100

pkg/controller/session/model_convert_test.go
bartoszmajsak: Right, as it's mainly conversion logic being tested I can think we are not really improving readability by a lot with moving/duplicating that.

cmd/ike/session/session_test.go
bartoszmajsak: Done in 87886dd

cmd/ike/cmd/develop.go
bartoszmajsak: Done in de3556f

PR #105, opened 2019-04-18
feat: enables to run ike against particular namespace

### LINKED ISSUE #103: Ability to specify a namespace we want to develop against
Currently, we rely on default namespace being defined in kube context. This, however, limits our testing possibilities, as we might want to use `ike` for two different namespaces at the same time.

Since telepresence allows to specify `--namespace` as an argument we should add corresponding flag and delegate.

bartoszmajsak:
Fixes #103

### DIFF
--- a/cmd/ike/session/session.go
+++ b/cmd/ike/session/session.go
@@ -49,7 +49,13 @@ func CreateOrJoinHandler(cmd *cobra.Command) (func(), error) {
 	route, _ := cmd.Flags().GetString("route")
 	sessionName := getSessionName(cmd)
 
-	h := &handler{c: DefaultClient(namespace),
+	client, err := DefaultClient(namespace)
+
+	if err != nil {
+		return func() {}, err
+	}
+
+	h := &handler{c: client,
 		ref:         deploymentName,
 		namespace:   namespace,
 		route:       route,
--- a/cmd/ike/session/session.go
+++ b/cmd/ike/session/session.go
@@ -74,7 +74,7 @@ func (h *handler) createOrJoinSession() (string, error) {
 		}
 		return h.waitForRefToComplete()
 	}
-	// join sessionName
+	// join session
 	session.Spec.Refs = append(session.Spec.Refs, h.ref)
 	err = h.c.Create(session)
 	if err != nil {
--- a/cmd/ike/session/session.go
+++ b/cmd/ike/session/session.go
@@ -15,7 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-// Handler is a function to setup a server session before attempting to connect. Returns a 'cleanup' function.
+// Handler is a function to setup a server sessionName before attempting to connect. Returns a 'cleanup' function.
 type Handler func(cmd *cobra.Command) (func(), error)
 
 // Offline is a empty Handler doing nothing. Used for testing
@@ -23,17 +23,33 @@ func Offline(cmd *cobra.Command) (func(), error) {
 	return func() {}, nil
 }
 
-// CreateOrJoinHandler provides the option to either create a new session if non exist or join an existing.
+type handler struct {
+	c           *client
+	ref,
+	namespace,
+	sessionName,
+	route       string
+}
+
+// CreateOrJoinHandler provides the option to either create a new sessionName if non exist or join an existing.
 // Rely on the following flags:
+//  * namespace - the name of the target namespace where deployment is defined
 //  * deployment - the name of the target deployment and will update the flag with the new deployment name
-//  * session - the name of the session
+//  * sessionName - the name of the sessionName
 //  * route - the definition of traffic routing
 func CreateOrJoinHandler(cmd *cobra.Command) (func(), error) {
-	sessionName := getSessionName(cmd)
 	deploymentName, _ := cmd.Flags().GetString("deployment")
+	namespace, _ := cmd.Flags().GetString("namespace")
 	route, _ := cmd.Flags().GetString("route")
+	sessionName := getSessionName(cmd)
 
-	serviceName, err := createOrJoinSession(sessionName, route, deploymentName)
+	h := &handler{c: DefaultClient(namespace),
+		ref:         deploymentName,
+		namespace:   namespace,
+		route:       route,
+		sessionName: sessionName}
+
+	serviceName, err := h.createOrJoinSession()
 	if err != nil {
 		return func() {}, err
 	}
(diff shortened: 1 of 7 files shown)


aslakknutsen:
Is this not missing the namespace being used by the session_client? TP would run against different namespaces, but the session would never be made in the correct namespace so TP wouldn't find the D/DC?

bartoszmajsak:
> Is this not missing the namespace being used by the session_client? TP would run against different namespaces, but the session would never be made in the correct namespace so TP wouldn't find the D/DC?

You are totally right... not sure how I missed that

bartoszmajsak:
@aslakknutsen fixed with de8a5f7

cmd/ike/session/session.go
```diff
@@ -143,26 +135,26 @@ func waitForRefToComplete(sessionName, ref string) (string, error) {
...
+		_ = h.c.Delete(session)
 	} else {
-		_ = DefaultClient().Create(session)
+		_ = h.c.Create(session)
 	}
 }
 
 func getSessionName(cmd *cobra.Command) string {
-	sessionName, err := cmd.Flags().GetString("session")
+	sessionName, err := cmd.Flags().GetString("sessionName")
```
aslakknutsen: That this passes our test suite is a little scary: https://github.com/aslakknutsen/istio-workspace/blob/de8a5f7c4c09e8e98349ab1d59e72b52ec3e9427/cmd/ike/cmd/develop.go#L79

cmd/ike/session/session.go
```diff
@@ -143,26 +135,26 @@ func waitForRefToComplete(sessionName, ref string) (string, error) {
...
 
-func removeOrLeaveSession(sessionName, ref string) {
-	session, err := DefaultClient().Get(sessionName)
+func (h *handler) removeOrLeaveSession() {
+	session, err := h.c.Get(h.sessionName)
 	if err != nil {
 		return // assume missing, nothing to clean?
 	}
-	// more then one participant, update session
+	// more then one participant, update sessionName
```
aslakknutsen: search-replace to wide. makes no sense.

cmd/ike/session/session.go
```diff
@@ -43,32 +59,32 @@ func CreateOrJoinHandler(cmd *cobra.Command) (func(), error) {
...
 			return "", err
 		}
-		return waitForRefToComplete(sessionName, ref)
+		return h.waitForRefToComplete()
 	}
-	// join session
-
-	session.Spec.Refs = append(session.Spec.Refs, ref)
-	err = DefaultClient().Create(session)
+	// join sessionName
```
aslakknutsen: search-replace to wide. makes no sense.

cmd/ike/session/session.go
```diff
@@ -15,25 +15,41 @@ import (
...
+	sessionName,
+	route       string
+}
+
+// CreateOrJoinHandler provides the option to either create a new sessionName if non exist or join an existing.
 // Rely on the following flags:
+//  * namespace - the name of the target namespace where deployment is defined
 //  * deployment - the name of the target deployment and will update the flag with the new deployment name
-//  * session - the name of the session
+//  * sessionName - the name of the sessionName
```
aslakknutsen: search-replace to wide. makes no sense.

cmd/ike/session/session.go
```diff
@@ -15,25 +15,41 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-// Handler is a function to setup a server session before attempting to connect. Returns a 'cleanup' function.
+// Handler is a function to setup a server sessionName before attempting to connect. Returns a 'cleanup' function.
```
aslakknutsen: search-replace to wide. makes no sense.

cmd/ike/session/session_client_test.go
```diff
@@ -20,7 +20,7 @@ var _ = Describe("Session Client operations", func() {
 			Kind:       "Session",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "sample-session",
+			Name: "sample-sessionName",
```
aslakknutsen: search-replace to wide. makes no sense.

cmd/ike/session/session_client_test.go
```diff
@@ -34,11 +34,11 @@ var _ = Describe("Session Client operations", func() {
 		fakeClient := testclient.NewSimpleClientset()
 		client, _ := session.NewClient(fakeClient, "test-namespace")
 
-		It("should get created session byt its name", func() {
+		It("should get created sessionName byt its name", func() {
```
aslakknutsen: search-replace to wide. makes no sense.

cmd/ike/session/session_client_test.go
```diff
@@ -56,11 +56,11 @@ var _ = Describe("Session Client operations", func() {
 			Expect(creationErr).ToNot(HaveOccurred())
 		})
 
-		It("should delete session byt its name", func() {
+		It("should delete sessionName byt its name", func() {
```
aslakknutsen: search-replace to wide. makes no sense.

cmd/ike/session/session.go
bartoszmajsak: Lemme have a look :) but yeah, it is

cmd/ike/session/session.go
bartoszmajsak:
```suggestion
	// join session
```

cmd/ike/session/session.go
bartoszmajsak:
```suggestion
// Handler is a function to setup a server sessionName before attempting to connect. Returns a 'cleanup' function.
```

cmd/ike/session/session_client_test.go
bartoszmajsak:
```suggestion
It("should get created sessionName byt its name"
```

cmd/ike/session/session.go
bartoszmajsak:
```suggestion
	sessionName, err := cmd.Flags().GetString("session")
```

cmd/ike/session/session.go
bartoszmajsak:
```suggestion
//  * session - the name of the session
```

cmd/ike/session/session_client_test.go
bartoszmajsak:
```suggestion
		It("should delete session by its name", func() {
```

cmd/ike/session/session_client_test.go
bartoszmajsak:
```suggestion
			Name: "sample-session",
```

cmd/ike/session/session.go
bartoszmajsak:
```suggestion
	// more than one participant, update session
```

cmd/ike/session/session.go
bartoszmajsak: It seems we don't cover a case when we want to join an existing session - so even with this regression we anyway always created a new one. 

Joining existing would be a bit tricky with starting 2 TP instances, but I can provide a test where I create Session CR upfront and join it.

aslakknutsen: Hmm, shouldn't we technically end in a "err" here? https://github.com/aslakknutsen/istio-workspace/blob/36cdc4daddd549aa9577dac27cc253d77de8dc24/cmd/ike/session/session.go#L158 

Or is that only on type parse error?

bartoszmajsak: Yeah, that too. I already changed that.

About joining an existing session - maybe we could provide an option to persist a session, as currently when we stop `ike develop` we always clean up. How about opening an issue for that and ship the test together with it?

aslakknutsen: Hmm.. That would leave a Route in error when we tear down the Local service it points to. We could for the sake of test, or in prep for a future feature of 'the monitoring option/build image' replacement type of thing... 

Maybe just leave it as to why we need this? https://github.com/aslakknutsen/istio-workspace/issues/59 :)

bartoszmajsak: > That would leave a Route in error when we tear down the Local service it points to.

Correct - we would need to check if there are any "active" participants first, before routing. This also opens up a question on how to handle two participants working on the same service anyway (from the session concept point of view). Something we should discuss in depth I believe

bartoszmajsak: I added logging, as this is parse error most likely. In such a case, I think we should roll back to random session name.

cmd/ike/session/session_client.go
```diff
@@ -1,7 +1,7 @@
 package session
 
 import (
-	"log"
+	logger "log"
```
aslakknutsen: Different logger?

bartoszmajsak: One is on the client side and the other within the cluster. We can use the same though, won't harm

bartoszmajsak: Or is it...

bartoszmajsak: Fixed with #fd2c88f

bartoszmajsak: and bb11168

PR #106, opened 2019-04-24
fix(spi): add contract between develop cmd and session handler outsid…

aslakknutsen:
…e of cobra

### DIFF
--- a/cmd/ike/cmd/develop.go
+++ b/cmd/ike/cmd/develop.go
@@ -38,7 +38,9 @@ func NewDevelopCmd() *cobra.Command {
 			defer sessionClose()
 
 			// HACK: need contract with TP cmd?
-			cmd.Flags().Set("deployment", sessionState.DeploymentName)
+			if err := cmd.Flags().Set("deployment", sessionState.DeploymentName); err != nil {
+				return err
+			}
 
 			if err := build(cmd); err != nil {
 				return err
--- a/cmd/ike/cmd/develop.go
+++ b/cmd/ike/cmd/develop.go
@@ -31,11 +31,14 @@ func NewDevelopCmd() *cobra.Command {
 		},
 		RunE: func(cmd *cobra.Command, args []string) error { //nolint[:unparam]
 
-			sessionClose, err := sessions(cmd)
+			sessionState, sessionClose, err := sessions(cmd)
 			if err != nil {
 				return err
 			}
 			defer sessionClose()
+
+			// HACK: need contract with TP cmd?
+			cmd.Flags().Set("deployment", sessionState.DeploymentName)
 
 			if err := build(cmd); err != nil {
 				return err
--- a/cmd/ike/cmd/session_func_test.go
+++ b/cmd/ike/cmd/session_func_test.go
@@ -1,0 +1,82 @@
+package cmd_test
+
+import (
+	. "github.com/aslakknutsen/istio-workspace/cmd/ike/cmd"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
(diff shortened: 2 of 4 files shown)


cmd/ike/session/session.go
```diff
@@ -10,41 +11,48 @@ import (
...
 
-	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+// Options holds the variables used by the Session Handler
+type Options struct {
+	DeploymentName string // name of the initial resource to target
+	SessionName    string // name of the session create or join if exist
```
bartoszmajsak:
```suggestion
	SessionName    string // name of the session to create or join if exists
```

cmd/ike/cmd/develop.go
```diff
@@ -31,12 +31,15 @@ func NewDevelopCmd() *cobra.Command {
...
 
-			sessionClose, err := sessions(cmd)
+			sessionState, sessionClose, err := sessions(cmd)
 			if err != nil {
 				return err
 			}
 			defer sessionClose()
 
+			// HACK: need contract with TP cmd?
+			cmd.Flags().Set("deployment", sessionState.DeploymentName)
```
bartoszmajsak: Another possible `HACK` could be to create a shared variable (this is a single process anyway) and mutate that one instead of relying on the internals of Cobra. So something like `TelepresenceOptions` where we store what to pass and (maybe) how to map it to the CLI call. WDYT?

bartoszmajsak:
image

cmd/ike/cmd/develop.go
bartoszmajsak: Ok this sucks even more

aslakknutsen:
No idea what you're talking about? https://github.com/aslakknutsen/istio-workspace/pull/106/commits/8f811547e70fb85b1e6831ac0af4d655e733c9c3

cmd/ike/cmd/session_func_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+		var command *cobra.Command
+
+		BeforeEach(func() {
+			command = NewDevelopCmd()
+		})
+
+		It("should fail if namespace is not defined", func() {
+			_, err := ToOptions(removeFlagFromSet(command.Flags(), "namespace"))
+			Expect(err).To(HaveOccurred())
```
bartoszmajsak: Do we want to test specific error msgs or we are good with such checks?

cmd/ike/cmd/session_func_test.go
```diff
@@ -0,0 +1,96 @@
...
+			_, err := ToOptions(removeFlagFromSet(command.Flags(), "session"))
+			Expect(err).To(HaveOccurred())
+		})
+
+		It("should fail if route is not defined", func() {
+			_, err := ToOptions(removeFlagFromSet(command.Flags(), "route"))
+			Expect(err).To(HaveOccurred())
+		})
+
+		It("should convert namespace if set", func() {
```
bartoszmajsak: How about moving it to other `Context`? Those are already testing conversion logic, not preconditions, as the current context they belong to ("checking required command arguments") might suggest.

cmd/ike/cmd/session_func_test.go
aslakknutsen: https://github.com/aslakknutsen/istio-workspace/pull/107

cmd/ike/cmd/session_func_test.go
aslakknutsen: https://github.com/aslakknutsen/istio-workspace/pull/107

PR #110, opened 2019-05-02
fix: uses session name for routing

### LINKED ISSUE #58: Add dynamic DestinationRule Destination Subset config
Currently, hardcoded to "v1-test".

Possibly rely on session name as a suffix to the target v.

Affects `mutateDestinationRule` and `mutateVirtualService`

bartoszmajsak:
Fixes #58 

Bonus: godoc and var names cleanup

### DIFF
--- a/pkg/controller/session/model_convert.go
+++ b/pkg/controller/session/model_convert.go
@@ -7,7 +7,7 @@ import (
 
 const (
 	// DefaultRouteHeaderName holds the name of the Header used to route traffic if no Route is provided
-	DefaultRouteHeaderName = "X-Workspace-Route"
+	DefaultRouteHeaderName = "x-workspace-route"
 
 	// RouteStrategyHeader holds the Route Type keyword for a Header based Route strategy
 	RouteStrategyHeader = "header"
--- a/pkg/istio/destinationrule.go
+++ b/pkg/istio/destinationrule.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	// DestinationRuleKind is the k8 Kind for a istio DestinationRule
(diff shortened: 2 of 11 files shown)


pkg/controller/session/model_convert.go
```diff
@@ -7,7 +7,7 @@ import (
 
 const (
 	// DefaultRouteHeaderName holds the name of the Header used to route traffic if no Route is provided
-	DefaultRouteHeaderName = "X-Workspace-Route"
+	DefaultRouteHeaderName = "x-workspace-route"
```
bartoszmajsak: According to istio docs header matchers should be lower case

> The header keys must be lowercase and use hyphen as the separator, e.g. `x-request-id`

bartoszmajsak: https://istio.io/docs/reference/config/networking/v1alpha3/virtual-service/#HTTPMatchRequest

pkg/istio/destinationrule.go
```diff
@@ -64,7 +64,7 @@ func DestinationRuleRevertor(ctx model.SessionContext, ref *model.Ref) error { /
 		}
 
 		ctx.Log.Info("Found DestinationRule", "name", resource.Name)
-		mutatedDr, err := revertDestinationRule(*dr)
+		mutatedDr, err := revertDestinationRule(*dr, ctx)
```
aslakknutsen: Why is revert taking 'ctx' while mutate just the 'name'?

bartoszmajsak: Missed that, the code was changed multiple times where I was fiddling with the `ctx` to pass more metadata. That's clearly leftover. Fixed with 593618c

PR #117, opened 2019-05-15
docs: first cut of documentation

### LINKED ISSUE #88: First cut of documentation
- [ ] overview / vision
- [ ] use cases
- [ ] user guide
- [ ] development guide

bartoszmajsak:
Fixes #88


docs/modules/ROOT/pages/cli_reference.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+Below you can find a documented list of all commands and their available flags.
+
+== Available commands
+
+[#ike]
+=== `ike`
+
+Main command, which starts the operator.
+// TODO brief mentionn about operator and how it works + link to the dedicated docs
```
bartoszmajsak: @aslakknutsen you can use

```
 ```suggestion
 \```
```

block to fill-in this gap. But this equally well can go with the separated PR.

bartoszmajsak:
@aslakknutsen Just realized we still need the cluster-side of getting started. :D

Which makes me think - shouldn't we have `ike install` for the ease of user experience vs a bit crufty Makefile which no one will care about when they only want to use the tool?

aslakknutsen:
@bartoszmajsak Yea, I was actually thinking the same the other day but never got around to writing up an issue. I'm sure there will be many ways of install, e..g Market place etc, but it's an easy local dev/test setup option. Want to create an issue?

bartoszmajsak:
Here's the issue https://github.com/Maistra/istio-workspace/issues/126

PR #118, opened 2019-05-16
chore: allows to use strings for simple cmd

bartoszmajsak:
Instead of breaking down the command to separate string for each token we can now just pass a string. 

Shortcoming - this does not work for sub-commands, e.g. `docker login -u $(oc whoami)`.

#### Why

I am working on the docs and aiming for extracting cmd samples from tests directly to the docs so we can have them always up-to-date. That's a first step.


aslakknutsen:
Have you tried to execute the cmd in a "sh -c" wrapper. I believe that's what I did to get $(x) working earlier.

bartoszmajsak:
> Have you tried to execute the cmd in a "sh -c" wrapper. I believe that's what I did to get $(x) working earlier.

I did not, let me give it a try.

PR #127, opened 2019-05-27
feat(testing): adds goleak check to each suite and template

bartoszmajsak:

I've added goleak when testing watch functionality - helped me finding a few corner cases in this part.

We somehow forgot to add that across the whole test suite.

- adds `AfterSuite` in each `_suite_test.go` with calls to goleak detection
- enhances ginkgo template so that new suites will have that check right from the beginning

### DIFF
--- a/cmd/ike/session/session_suite_test.go
+++ b/cmd/ike/session/session_suite_test.go
@@ -3,6 +3,8 @@ package session_test
 import (
 	"testing"
 
+	"go.uber.org/goleak"
+
 	. "github.com/maistra/istio-workspace/test"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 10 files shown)


cmd/ike/session/session_suite_test.go
```diff
@@ -13,3 +15,10 @@ func TestSession(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecWithJUnitReporter(t, "Session Suite")
 }
+
+var _ = SynchronizedAfterSuite(func() {}, func() {
+	goleak.VerifyNone(GinkgoT(),
```
aslakknutsen: At this point, one can start to wonder about that one 'global' suite..

bartoszmajsak: The only problem is that you can only have one "Before/AfterSuite" hook per suite, so we would need to put all the logic there, does not matter if it's used by every test or not. A bit of a disadvantage.

Issue #38, opened 2019-02-22
Watch should support glob patterns for inclusions


bartoszmajsak:
Currently we only support directory paths to watch for changes. In addition, we should let users define glob patterns, so they can limit files to react on. 

- if only relative patterns (e.g. `src/main/**/*.java`) are defined, the watch is applied on the current directory
- if an absolute paths are defined the relative patterns should be applied on those instead

PR #129, opened 2019-05-29
feat(watch): takes .gitignore into consideration if present in the excluded path

### LINKED ISSUE #36: Watch exclusions - defaults
Ignore common project files which should not be watched:

- scm folders
- IDE files
- logs

### LINKED ISSUE #37: Watch exclusions - .gitignore
If  `.gitignore` is found in the folder `ike` is watching we should exclude the same paths from `watch`.

bartoszmajsak:

When watching a project there can be multiple files being changed behind the scenes besides the one which is edited by the user (e.g. IDE files). This can lead to excessive event handling and reloads. Most frequently such artifacts are excluded through `.gitignore` so handling those patterns as part of `watch` would help.

- checks if `.gitignore` exists in the base path and supports it in such a case
- `.gitignore`s in subdirectories are not recursively looked-up - that might make the code overly complicated to handle ~5% of use cases (percentage out of thin air, of course ;)
- adds `go-gitignore` library for handling `.gitignore` patterns

Fixes #36 #37 #38


bartoszmajsak:
- [ ] rework user-defined exclusion to use the same library as `.gitignore`
- [ ] document how `.gitignore` is handled (no sub dirs)

PR #130, opened 2019-05-30
feat: adds --version flag to root cmd

bartoszmajsak:

Among the cli tools you can find both `cmd --version` and `cmd version` way of getting information about the version of the tool in use. I frequently mix it up so I thought it would be good to actually have both available :)

- adds `--version` flag to root command which only prints version and skips operator start

### DIFF
--- a/cmd/ike/cmd/root.go
+++ b/cmd/ike/cmd/root.go
@@ -38,6 +38,10 @@ func NewRootCmd() *cobra.Command {
 		},
 		RunE: func(cmd *cobra.Command, args []string) error { //nolint[:unparam]
 			printVersion()
+			justPrintVersion, _ := cmd.Flags().GetBool("version")
+			if justPrintVersion {
+				return nil
+			}
 			return startOperator()
 		},
 	}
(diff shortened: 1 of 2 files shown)


cmd/ike/cmd/root.go
```diff
@@ -38,14 +38,18 @@ func NewRootCmd() *cobra.Command {
 		},
 		RunE: func(cmd *cobra.Command, args []string) error { //nolint[:unparam]
 			printVersion()
+			justPrintVersion, _ := cmd.Flags().GetBool("version")
```
aslakknutsen: Always print version, but you can choose to only print the version?

bartoszmajsak: I preserved original behaviour, as I thought it's helpful on the server side when the service starts.

bartoszmajsak: Alternatively, we can refactor this out and provide `serve` (or `start` or any better name) subcommand which will do operator thing in the cluster.

bartoszmajsak: Maybe having `develop` vs `serve` subcommand will make it more clear on what this tool provides (vs root which is used right now).

aslakknutsen: I think a 'serve' sub command would work

bartoszmajsak: I will create new issue then and provide separated PR. Now back to this change - are we ok with merging it or we stick to `ike version`?

bartoszmajsak: #131

PR #137, opened 2019-06-05
bug(clone): Remove fields that can cause problems for the cloned proxy

### LINKED ISSUE #135: Remove liveness/readiness probes from cloned Deployment/DeploymentConfig
The pod will fail to start with the telepresence proxy if `livenessProbe` and `readinessProbe` are defined on the original deployment.

aslakknutsen:
fixes #135

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -66,12 +66,12 @@
   version = "v2.9.6"
 
 [[projects]]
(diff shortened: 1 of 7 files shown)


bartoszmajsak:
Are we using `bug()` now or `fix()`? :)

Issue #142, opened 2019-06-06
Extract Env var references in test suite and code into a common const set


aslakknutsen:

bartoszmajsak:
I think it would be a good idea to store those in a map too with defaults as values, so we can iterate over them and print as part of help and documentation.

PR #145, opened 2019-06-10
feat: ike install-operator command for deploying the operator

### LINKED ISSUE #126: Installer - provide a command to deploy operator
##### What can be improved

By providing `ike install` command we can simplify the installation process for the end-user which will hide all the complexity.

We could bundle processed templates during the build of the ike binary for that purpose. The only thing here to think through would be our local development flow where we change the location of the docker images to internal cluster registry, but I guess we could simply rely on env variables and read them while running `ike install`.

##### Why

The operator installation process involves `Makefile` and several steps one has to perform before using `ike develop`. This is not an appealing option for regular users, especially considering the fact that we aim to ship simple binary. No one will be eager to clone the repo to run `Makefile`.

bartoszmajsak:

Provides an easy way of installing istio-workspace operator into the cluster through the self-contained binary.

- `ike install` which deploys operator to the cluster
- ability to specify namespace by using `--namespace` or `-n` flag
- tests installer logic in e2e test setup
- bundles deployment resources into the binary by using `go-bindata` library

Fixes #126


bartoszmajsak:
I am checking what's up with the build, as it obviously works on my machine.

bartoszmajsak:
@aslakknutsen I am just wondering if `install` is not misleading as we have two sides to cover - cluster (which this command provides) and client which requires TP to be present. 

Should we rename to `install-cluster` or provide a flag `--cluster` for later automation of the client-side prerequisites? WDYT?

aslakknutsen:
@bartoszmajsak install-operator? Did you sleep on the "install" vs "dump"?

bartoszmajsak:
I think `install-operator` is fine for the time being.

pkg/assets/isto-workspace-deploy.go
```diff
@@ -0,0 +1,365 @@
...
+	names := make([]string, 0, len(_bindata))
+	for name := range _bindata {
+		names = append(names, name)
+	}
+	return names
+}
+
+// _bindata is a table, holding each asset generator, mapped to its name.
+var _bindata = map[string]func() (*asset, error){
+	"deploy/istio-workspace/crds/istio_v1alpha1_session_cr.yaml":  deployIstioWorkspaceCrdsIstio_v1alpha1_session_crYaml,
```
aslakknutsen: This one is not needed and technically only for testing/example

aslakknutsen:
@bartoszmajsak Some form of feedback on what is going on when you execute the command would be helpful. Print each resource being created or similar. Just nothing makes you wonder if it failed or not.

pkg/assets/isto-workspace-deploy.go
```diff
@@ -0,0 +1,365 @@
+// Package assets Code generated by go-bindata. (@generated) DO NOT EDIT.
```
aslakknutsen: What's the difference between this and "pkg/assets/generated/isto-workspace-deploy.go"? They seem pretty dup

pkg/openshift/client.go
```diff
@@ -0,0 +1,120 @@
+package openshift
```
aslakknutsen: Any reason for keeping this separate from cmd/ike/session/session_client?

cmd/ike/cmd/install.go
```diff
@@ -0,0 +1,127 @@
...
+func installOperator(cmd *cobra.Command, args []string) error { //nolint[:unparam]
+	namespace, err := cmd.Flags().GetString("namespace")
+	if err != nil {
+		return err
+	}
+	app, err := newApplier(namespace)
+	if err != nil {
+		return err
+	}
+	if err := app.applyResource("crds/istio_v1alpha1_session_crd.yaml"); err != nil {
```
aslakknutsen: Technically all these could have been replaced with something like:

```go
	resources := []string{"crds/istio_v1alpha1_session_crd.yaml", "service_account.yaml"}
	for _, resource := range resources {
		if r, err := app.applyResource(resource); err != nil { // r = TypeMeta?
			if !errors.IsAlreadyExists(err) {
				log.Error(err, "failed creating " + r)
				return err
			}
			log.Info(r + " already exists")
		}
	}
```

pkg/openshift/client.go
bartoszmajsak: We can put this together. My thinking was that this one is a generic-purpose client which has no typed interfaces so `openshift` seemed to be a more natural place.

pkg/openshift/template.go
```diff
@@ -0,0 +1,127 @@
+package openshift
```
aslakknutsen: Should we separate between "pkg/openshift" the Operator stuff and the "pkg/openshift" 'Other stuff'?

aslakknutsen: maybe pkg/openshift/client ?

aslakknutsen: or move istio|k8s|openshift "operator stuff" into a common parent package..

pkg/openshift/template.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+var re = regexp.MustCompile(`(?m)\${(.*?)}`)
+
+const substitution = "{{ .$1 }}"
+
+// ProcessTemplateUsingEnvVars takes template path, loads its content and
+// substitutes variables by using those which are defined as environment variables.
+// Returns processed template as byte array
+func ProcessTemplateUsingEnvVars(templatePath string) ([]byte, error) {
```
aslakknutsen: Would have been nice if maybe as part of 'ike install-operator --help' it displayed which ENV variables were available to manipulate. A little dump of the Template.parameters. (note: there is a "description" field on the parameters as well we could use=

bartoszmajsak: Yeah, I'm working on it as part of #142

cmd/ike/cmd/install.go
aslakknutsen: as I wrote that I do now see there are applyResource and applyTemplate but.. :)

pkg/assets/isto-workspace-deploy.go
bartoszmajsak: Renamed file to `istio_v1alpha1_session_cr.example.yaml` and made `go-bindata` ignoring all example files. See 1616034

pkg/assets/isto-workspace-deploy.go
bartoszmajsak: Seems to me like leftover before tweaking go-bindata Make target. Fixed 72ac2c3

cmd/ike/cmd/install.go
bartoszmajsak: True, I felt this code is crap :) Fixed in 99a78b5

pkg/openshift/client.go
aslakknutsen: Was not thinking about the location, but rather function. Minimally we could use the 'non typed' 'client' internally within the 'SessionClient' interface. It's our Resource so we shouldn't really be surprised by changes to it. hehe

bartoszmajsak:
> @bartoszmajsak Some form of feedback on what is going on when you execute the command would be helpful. Print each resource being created or similar. Just nothing makes you wonder if it failed or not.

Done in db2da47

pkg/openshift/client.go
bartoszmajsak: Ah ok, got it. Let me go over the code and see - there's definitely some dupl in a way the client is set up for example. We can have both typed and generic in one place, why not.

pkg/openshift/template.go
bartoszmajsak: I did a small shuffling around. I opened #150 issue to clean up the client stuff.

pkg/openshift/client.go
bartoszmajsak: I opened #150 issue to clean up the client stuff.

PR #147, opened 2019-06-12
chore: minor code fixes from running goland analysis

bartoszmajsak:

Goland "Inspect Code" has some additional rules which our linter do not provide. I ran them sporadically to check if we have missed sth

- explicit ignore of errors in deferred `.Close()`
- removes unused variable `ikeBin`
- renames `for _, url := range config.Call` to `for _, u := range config.Call` to avoid colliding with imported `net/url` package name


cmd/ike/watch/watch_test.go
```diff
@@ -147,10 +147,10 @@ src/main/**/*.java
 		code := TmpFile(GinkgoT(), watchTmpDir+"/main.go", "package main")
 
 		defer func() {
-			config.Close()
-			test.Close()
-			nestedFile.Close()
-			gitIgnore.Close()
+			_ = config.Close()
```
aslakknutsen: print error instead of just completely ignore?

bartoszmajsak: That would be an improvement :P

PR #151, opened 2019-06-13
feat: track env vars from templates

### LINKED ISSUE #142: Extract Env var references in test suite and code into a common const set

bartoszmajsak:

Collects env vars/parameters defined in templates so we can document them easily in the CLI and 

- adds a description to parameters in templates
- extracts env var string to consts
- collect tpl parameters

Fixes #142 

NOTE: help string and actual usage will come as part of #117


cmd/ike/config/env_vars.go
```diff
@@ -0,0 +1,38 @@
...
+	EnvDockerRegistry   = "IKE_DOCKER_REGISTRY"
+)
+
+var (
+	log        = logf.Log.WithName("session_handler")
+	parameters []openshiftApi.Parameter
+)
+
+func init() {
+	collectTplParams("operator.yaml")
```
aslakknutsen: For future safety; iterate over all Assets?

pkg/openshift/parser/template_test.go
```diff
@@ -113,22 +113,27 @@ var _ = Describe("template processing", func() {
 	})
 })
 
-const processedTmplWithDefaults = `kind: Template
+const processedOperatorTmplWithDefaults = `kind: Template
```
aslakknutsen: Any reason to keep this vs reading the actual file?

test/cmd/test-scenario/main.go
```diff
@@ -37,10 +39,10 @@ func getTestImageName() string {
...
 
-	if s, f := os.LookupEnv("IKE_DOCKER_REGISTRY"); f {
+	if s, f := os.LookupEnv(config.EnvDockerRegistry); f {
 		reg = s + "/"
 	}
-	if s, f := os.LookupEnv("IKE_DOCKER_REPOSITORY"); f {
+	if s, f := os.LookupEnv(config.EnvDockerRepository); f {
 		repo = s
 	}
 	if s, f := os.LookupEnv("IKE_TEST_IMAGE_NAME"); f {
```
aslakknutsen: Add to config even tho not in Template?

pkg/openshift/parser/template_test.go
bartoszmajsak: This one is processed - so have values replaced.

test/cmd/test-scenario/main.go
bartoszmajsak: as `const`?

bartoszmajsak: 1177045

cmd/ike/config/env_vars.go
bartoszmajsak: Ace! ace0ee7

PR #154, opened 2019-06-14
feat(vs): do a better job at setting up the correct route

### LINKED ISSUE #153: Fine tune VirtualService mutation
* Remove mirror
* Remove weight
* Remove non targeted Destinations from route
* Remove Redirect
* Add headers to each existing match

aslakknutsen:
fixes #153

This requires some update to the API between Locators and Mutator/Revertors as Subset is missing.

### DIFF
--- a/pkg/istio/virtualservice.go
+++ b/pkg/istio/virtualservice.go
@@ -1,6 +1,7 @@
 package istio
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/maistra/istio-workspace/pkg/model"
@@ -35,7 +36,7 @@ func VirtualServiceMutator(ctx model.SessionContext, ref *model.Ref) error { //n
 	}
 
 	ctx.Log.Info("Found VirtualService", "name", targetName)
-	mutatedVs, err := mutateVirtualService(ctx, ref.Target.GetNewVersion(ctx.Name), *vs)
+	mutatedVs, err := mutateVirtualService(ctx, ref.Target, *vs)
 	if err != nil {
 		ref.AddResourceStatus(model.ResourceStatus{Kind: VirtualServiceKind, Name: targetName, Action: model.ActionFailed})
 		return err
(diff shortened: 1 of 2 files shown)


pkg/istio/virtualservice.go
```diff
@@ -83,51 +84,78 @@ func VirtualServiceRevertor(ctx model.SessionContext, ref *model.Ref) error { //
...
+	}
+	for _, targetHTTP := range targetsHTTP {
+		targetHTTP = removeOtherRoutes(targetHTTP, sourceResource.GetHostName(), sourceResource.GetVersion())
+		targetHTTP = updateSubset(targetHTTP, sourceResource.GetNewVersion(ctx.Name))
+		targetHTTP = addHeaderMatch(targetHTTP, ctx.Route)
+		targetHTTP = removeWeight(targetHTTP)
+		targetHTTP.Mirror = nil
+		targetHTTP.Redirect = nil
+
+		target.Spec.Http = append(target.Spec.Http, &targetHTTP)
```
golangcibot: Using a reference for the variable on range scope `targetHTTP` (from `scopelint`)

pkg/istio/virtualservice.go
```diff
@@ -83,51 +84,78 @@ func VirtualServiceRevertor(ctx model.SessionContext, ref *model.Ref) error { //
...
 
-	return vs, nil
+	target := source.DeepCopy()
+	clonedSource := source.DeepCopy()
+
+	targetsHTTP := findRoutes(clonedSource, sourceResource.GetHostName(), sourceResource.GetVersion())
+	if len(targetsHTTP) == 0 {
+		return istionetwork.VirtualService{}, fmt.Errorf("route not found")
+	}
+	for _, targetHTTP := range targetsHTTP {
```
golangcibot: rangeValCopy: each iteration copies 192 bytes (consider pointers or indexing) (from `gocritic`)

pkg/istio/virtualservice.go
```diff
@@ -83,51 +84,78 @@ func VirtualServiceRevertor(ctx model.SessionContext, ref *model.Ref) error { //
...
+		targetHTTP = removeOtherRoutes(targetHTTP, sourceResource.GetHostName(), sourceResource.GetVersion())
+		targetHTTP = updateSubset(targetHTTP, sourceResource.GetNewVersion(ctx.Name))
+		targetHTTP = addHeaderMatch(targetHTTP, ctx.Route)
+		targetHTTP = removeWeight(targetHTTP)
+		targetHTTP.Mirror = nil
+		targetHTTP.Redirect = nil
+
+		target.Spec.Http = append([]*v1alpha3.HTTPRoute{&targetHTTP}, target.Spec.Http...)
```
golangcibot: Using a reference for the variable on range scope `targetHTTP` (from `scopelint`)

bartoszmajsak:
As for e2e_test failure - it seems `telepresence` package is not reachable... 

`E: Unable to locate package telepresence`

bartoszmajsak:
We might want to think about the fallback plan to avoid such roadblocks.

aslakknutsen:
> We might want to think about the fallback plan to avoid such roadblocks.

Package cache?

aslakknutsen:
> I just wonder what do we benefit having funcs defined in the `mutateVirtualService` vs making them entities on their own. They seem not to rely on any struct passed to "owning func".

That was primarily done  for two reasons:

* to make the changes clear 'at the bottom of the method' as opposed to a long bunch of loops etc
* and kept in inside the func to indicate they are not intended to be reusable and not to "polute" the namespace.

bartoszmajsak:
> > We might want to think about the fallback plan to avoid such roadblocks.
> 
> Package cache?

Opened an issue for that one #201

PR #158, opened 2019-06-20
feat(circle): disables linter for ci build

### LINKED ISSUE #157: Introduce GolangCI service
##### What can be improved

Use of `golangci` linter as a service.

##### Why

Currently, we ran the full scan of `golangci` as part of the build in `Makefile`. We can speed it up by scanning only a range of changes and relying on external SaaS-based `golangci` instance which is free for open source projects. Additionally, it will make CircleCI builds faster.

bartoszmajsak:

As we have now golangci as separated PR check this will disable linter for CI build on CircleCI.

- explicitly defines the default goal (all)
- introduces `build-ci` target without linter
- switches to `build-ci` for circle-ci builds
- first `compiles` then run `tests` for `all` and `build-ci`

Fixes #157

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
           name: "Runs the build"
           command: |
             make tools && \
-            make
+            make build-ci
       - save_cache:
           key: vendor-cache-{{ checksum "Gopkg.lock" }}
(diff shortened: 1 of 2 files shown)


bartoszmajsak:

PR #162, opened 2019-06-21
chore: updates Telepresence to latest 0.101

bartoszmajsak:

There's a new version of Telepresence (0.101) out, so we should bump that in our tool.

- default TP version to 0.101
- updates operator template


bartoszmajsak:

PR #165, opened 2019-06-21
chore(make): extracts version from git tag

bartoszmajsak:
Ongoing version in the built binary will be `latest-tag-next`, so that we know from which stable release it evolved.


bartoszmajsak:

PR #171, opened 2019-06-28
fix: versioning for operator installer

bartoszmajsak:

Simplifies versioning and re-use the version in the operator installation process.

- cleaned up versioning scheme
- docker image tag is same as `VERSION` now
- makes `docker-build` rely on `compile` to always build latest binary
- `ike install-operator` uses compile-time defined `version.Version` param to use for template processing if `IKE_IMAGE_TAG` is not defined as env var

#### Versioning scheme
Following versioning scheme is proposed. Let's take the following sequence of commits:

`--> a4f --> b5d (tag: v0.0.1) --> cdf --> d56 (tag: v0.0.2) --> 4cb --> ...`

- binary built against `b5d` will have a version `v0.0.1` and it will push docker image tagged with the same
- if there are changes on this commit subsequent builds will be labeled `v0.0.1-dirty-TIMESTAMP`
- build on `cdf` will have `v0.0.1-next` prefix with the same dirty rules as above
- build on `4cb` will have `v0.0.2-next` prefix with the same dirty rules as above

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -2,7 +2,7 @@ package version
 
 var (
 	// Version hold a semantic version of the running binary
-	Version = "0.0.0"
+	Version = "v0.0.0"
 	// Commit holds the commit hash against which the binary build was ran
 	Commit string
 	// BuildTime holds timestamp when the binary build was ran
--- a/version/version.go
+++ b/version/version.go
@@ -2,7 +2,7 @@ package version
 
 var (
 	// Version hold a semantic version of the running binary
-	Version = "0.0.1"
+	Version = "0.0.0"
 	// Commit holds the commit hash against which the binary build was ran
 	Commit string
 	// BuildTime holds timestamp when the binary build was ran
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -15,7 +15,7 @@ builds:
     - 386
     - amd64
   ldflags:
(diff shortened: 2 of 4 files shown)


version/version.go
```diff
@@ -2,7 +2,7 @@ package version
 
 var (
 	// Version hold a semantic version of the running binary
-	Version = "0.0.1"
+	Version = "0.0.0"
```
aslakknutsen: When run via Make the Version will be in format "v0.0.0"

aslakknutsen:
`docker-push` is left kinda hanging. required to run "docker-build docker-push" in 1 go. But not to bad.

But was wondering if maybe we should split docker-push or somehow make docker-push :latest only happen when a clean v tag is used? To avoid pushing local builds or unreleased stuff as latest ?

version/version.go
aslakknutsen:
```suggestion
	Version = "v0.0.0"
```

bartoszmajsak:
```suggestion
	Version = "v0.0.0"
```

bartoszmajsak:
> But was wondering if maybe we should split docker-push or somehow make `docker-push` :latest only happens when a clean v tag is used? To avoid pushing local builds or unreleased stuff as latest?

Lemme work on it once we release `v0.0.1` :)

Issue #164, opened 2019-06-21
Publish docker images


bartoszmajsak:
##### What can be improved

Automated release of operator docker images

##### Why

Repetitive, boring tasks should be automated

bartoszmajsak:
Ultimately went with `Makefile` approach which we already use regardless.

PR #172, opened 2019-07-01
feat(build): circle release build automation

### LINKED ISSUE #163: Release automation as CI job
##### What can be improved

Instead of running it from dev machine we can have it as part of CircleCI.

The job would react on the pushed tag.

##### Why

image

#### How

```yaml
workflows:
  version: 2
  build-n-release:
    jobs:
      - build:
          filters:  # required since `release` has tag filters AND requires `build`
            tags:
              only: /.*/
      - release:
          requires:
            - build
          filters:
            tags:
              only: /^v.*/
            branches:
              ignore: /.*/
```

* the build job runs for all branches and all tags.
* runs release job on a tag and tag only using dedicated `filter` - tags starting with `v`
* release job runs goreleaser logic

bartoszmajsak:

Release automation based on a tagged commit (named `release: v0.0.1`) with the following content:

1. The version in `docs/antora.yml` changed to matching `tag`
2. Release notes named after tag `tag.adoc` in `docs/modules/ROOT/pages/release_notes/`
3. New release included in `docs/modules/ROOT/pages/release_notes.adoc`

The after-release commit should revert `docs/antora.yml` version to `latest`.

NOTE: provided as `scripts/release.sh`

- splits Make target for `docker-push` for `-latest` and `-versioned`
- changes docker registry to `quay.io`
- enhances circleci build setup with new `release` containing:
  - trigger on tags
  - pushing docker image for tag
  - `goreleaser` to release binaries and associated release notes
- release notes for `v0.0.1`

Fixes #163 #164

#### Pre-merge steps

- [ ] populate `quay.io` credentials in circle-ci (`QUAY_AUTH_JSON`)
- [ ] add Github personal token for goreleaser to push from circle-ci

#### Post-merge steps

- [x] tag and push
  - change `antora.yaml` version to match created tag
- [x] revert `antora.yaml` version to `latest`


bartoszmajsak:
Automation flows

bartoszmajsak:
Here's created release with custom notes https://github.com/bartoszmajsak/istio-workspace/releases/tag/v0.0.1

bartoszmajsak:
Built images can be found here https://quay.io/repository/bmajsak/istio-workspace?tab=tags

.circleci/config.yml
```diff
@@ -81,6 +81,42 @@ jobs:
...
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+      - checkout
+      - run:
+          name: "Sets up Quay.io"
+          command: |
+            mkdir -p ~/.docker && chmod 0700 ~/.docker
+            touch ~/.docker/config.json && chmod 0600 ~/.docker/config.json
+            echo $QUAY_AUTH_JSON > ~/.docker/config.json
```
bartoszmajsak: As long as we have this option turned off we are good with this approach (and in fact we do have it disabled! \o/)
image

bartoszmajsak:
@aslakknutsen I was thinking about the follow-up task for publishing `latest` docker images on every successful build on `master`, but in this case, we should have a way to publish snapshot binary on its own. Could be done by "moving tag" on GH, but I don't like this idea a lot. So I wonder do we need that at all?

bartoszmajsak:
If you like the image about I can add that to the docs ;)

aslakknutsen:
Nice! You do the pre/post steps?

bartoszmajsak:
> Nice! You do the pre/post steps?

Yeah, adding the script in a bit.

bartoszmajsak:
@aslakknutsen I tried to automate adding release notes to the `.adoc` file by using glob pattern, but it seems antora doesn't have an easy way to scan directories during build (all what I get is relative path to the docs project, which is not what we need). I tried to port https://github.com/asciidoctor/asciidoctor-extensions-lab/tree/master/lib/glob-include-processor to js.

So the script shipped with bf57c3f will simply append .adoc file based on the tag you provide.

aslakknutsen:
> @aslakknutsen I was thinking about the follow-up task for publishing latest docker images on every successful build on master, but in this case, we should have a way to publish snapshot binary on its own. Could be done by "moving tag" on GH, but I don't like this idea a lot. So I wonder do we need that at all?

Not sure if Latest means the latest commit or latest release. I would say release. So a push to :latest should be part of the release script. Not to say we couldn't have a :snapshot or similar as well on master builds. But as far as snapshot client storage, good q. Script to extract from Container image? hehe

PR #177, opened 2019-07-03
feat: download script for ike binary

### LINKED ISSUE #116: Binary installation simplification
Once we get the binary out in the wild we should provide simple means of distributing/installing it.

- [ ] shell script (`curl -sSL https://git.io/xxxxx | bash`)
- [ ] go get

bartoszmajsak:

`get.sh` script (inspired by [goreleaser](https://github.com/goreleaser/get/blob/03c4bfde763b30bfe270892ab3ff74949d8e9351/get)). It downloads the binary to the and hints how to add it 

It's purposefully called `get.sh` (which is somewhat generic) instead of `download.sh`, because if we evolve it to the actual installer we still can keep `curl -sL http://git.io/get-ike | bash` snippet. The URL has been created for `get.sh` and there's no way to overwrite it in the future.

- general purpose `get.sh` script which will download latest version of the binary for the given system
- update of docs
- update of README

Fixes #116

### DIFF
--- a/scripts/get.sh
+++ b/scripts/get.sh
@@ -68,7 +68,7 @@ while test $# -gt 0; do
             if test $# -gt 0; then
               dir=$1
             else
-              die "Please provide a version name"
+              die "Please provide a version"
             fi
             shift
             ;;
--- a/scripts/get.sh
+++ b/scripts/get.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+## Derived from https://github.com/goreleaser/get/blob/03c4bfde763b30bfe270892ab3ff74949d8e9351/get
+
+set -e
+
+die () {
+    echo >&2 "$@"
+    exit 1
+}
+
+TAR_FILE="/tmp/ike.tar.gz"
+RELEASES_URL="https://github.com/maistra/istio-workspace/releases"
+
+last_version() {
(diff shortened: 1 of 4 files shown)


scripts/get.sh
```diff
@@ -0,0 +1,38 @@
...
+test -z "$VERSION" && {
+  die "Unable to get ike version. You can still try to download manually from $RELEASES_URL"
+}
+test -z "$TMPDIR" && TMPDIR="$(mktemp -d --suffix=-ike-${VERSION})"
+
+download
+tar -xf "$TAR_FILE" -C "$TMPDIR"
+
+echo "Downloaded ike binary to $TMPDIR"
+echo -e "You can add it to your path by typing following in your terminal:\n$ export PATH=\"\$PATH:$TMPDIR\""
```
aslakknutsen: TMPDIR is deleted on reboot etc isn't it?

bartoszmajsak: Right. Should add a flag or variable so that the user can define the location himself?

aslakknutsen: yea.. the alternative is to require root and place it in /usr/local/bin or something...

bartoszmajsak: e424305 brings the following behaviour:

```
╰─ cat scripts/get.sh | bash -s -- --dir=./dist
Downloaded ike binary (v0.0.1) to ./dist
Make sure it's on your $PATH.
```

Good enough?

PR #180, opened 2019-07-05
chore: moves cmd logic to pkg/

bartoszmajsak:

`cmd/` should only contain entrypoint to the commands we provide (i.e. `main.go` and no other files and folders). `pkg/cmd` should consist of the actual logic for commands themselves.

- moves logic to `pkg/`
- `session` and `watch` made `internal`


aslakknutsen:
We're considering the pkg/cmd as reusable commands by others?

bartoszmajsak:
> We're considering the pkg/cmd as reusable commands by others?

Dunno - should we? If not then where should it live?

aslakknutsen:
If they are not intended to be, should they not be in internal?

bartoszmajsak:
> If they are not intended to be, should they not be in internal?

Sure, except then we can't use it in `cmd/ike/main.go` :)

aslakknutsen:
Why not? We can do whatever we want. It's more about what 'other external' should or shouldn't do.

bartoszmajsak:
> Why not? We can do whatever we want. It's more about what 'other external' should or shouldn't do.

It's not just making things `internal` - requires restructuring the code. But it, in fact, might end up in the better structure. On it.

bartoszmajsak:
Here's the new structure brought by 14bc3b1

```
├── apis
│   └── istio
│       └── v1alpha1
├── assets
├── client
│   └── clientset
│       └── versioned
│           ├── fake
│           ├── scheme
│           └── typed
│               └── istio
├── cmd
│   ├── config
│   ├── develop
│   │   └── internal
│   ├── format
│   ├── install
│   ├── internal
│   │   └── build
│   ├── serve
│   ├── version
│   ├── watch
│   └── root.go
├── controller
│   └── session
├── internal
│   └── session
├── istio
├── k8s
├── model
├── naming
├── openshift
│   ├── dynclient
│   └── parser
├── shell
└── watch
```

aslakknutsen:
@bartoszmajsak Might be needed by something like 'ike list sessions' to see which to join? Auto complete?

bartoszmajsak:
> @bartoszmajsak Might be needed by something like 'ike list sessions' to see which to join? Auto complete?

Then let's leave it as is

PR #181, opened 2019-07-08
feat: ike bash and zsh autocomplete

### LINKED ISSUE #17: Provide bash and zsh autocompletion

bartoszmajsak:

Provides CLI completion for `bash` and `zsh`.

```shell
. <(ike completion zsh)
```

demo

- shows completion of required flags first, and only when they are provided the rest is hinted
- delegates completion of certain objects (i.e. `namespace`) to underlying tooling (`kubectl` in this case)
- namespace-aware completion: whenever namespace has been specified before any namespace-bound parameter (such as `--deployment`), we are narrowing the results to that specific namespace (see `__kubectl_get`)
- combining object sets: we can combine results of multiple `kubectl get`s by calling `__kubectl_get_object_combined` with multiple types combined - this is in particular useful for `--deployment` used by `telepresence` which accepts both `deployment` from k8s and `deploymentconfig` from Openshift

Fixes #17


docs/modules/ROOT/pages/cli_reference.adoc
```diff
@@ -5,6 +5,11 @@ In this section you will learn about:
 * [x] ways to configure `ike` (through flags, files et al)
 * [x] complete list of commands and their available flags
 
+[#completion]
+== Autocomplete
+
+If you are using `zsh` or `bash` you can easily enable autocomplete for `ike` by invoking `. <(ike completion SHELL)`.
+
```
aslakknutsen: Should this not have a section like:

   include::cmd:ike[flags='completion --help --help-format=adoc']

bartoszmajsak: Not sure if that would be better

bartoszmajsak: I mean this is one-off not changing stuff, but in general, I'm thinking of extending help generation to have not only flags output, but also other bits in the template, include usage and example if exists.

On top of that iterating over all the commands instead of having section added manually.

I've opened issue for that #182. Happy to work on it next.

aslakknutsen: Didn't mean instead of. Just thinking if we explain "all" commands in one place in one way we might as well explain all.

bartoszmajsak: We don't yet - we just print flags.

bartoszmajsak:
#### Tasks

- [ ] few tests for completion

docs/modules/ROOT/pages/cli_reference.adoc
aslakknutsen: Oh sure, I just meant under "== Available commands"

bartoszmajsak: #182

e2e/completion/bash_completion_test.go
```diff
@@ -0,0 +1,132 @@
...
+		})
+
+		Context("install-operator", func() {
+
+			It("should show matching command", func() {
+				Expect(completionFor("ike install-")).To(ConsistOf("install-operator"))
+			})
+
+			It("should show all flags inherited from root", func() {
+				Expect(completionFor("ike install-operator -")).To(ConsistOf("--config", "--config=","-c", "--namespace", "--namespace=", "-n"))
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
				Expect(completionFor("ike install-operator -")).To(ConsistOf("--config", "--config=", "-c", "--namespace", "--namespace=", "-n"))
```

aslakknutsen:
ike install-operator <tab>

Expected: Show all options

Actual: Shows non, but assumes a 'file' input. -<tab> cause other flags to be completed

bartoszmajsak:
> Expected: Show all options
> 
> Actual: Shows non, but assumes a 'file' input. -<tab> cause other flags to be completed

can you check if you can observe the same behaviour for `kubectl` or `oc` completions? Wondering if it's something in Cobra or our stuff

aslakknutsen:
oc has the same behavior. Seems you need to 'start' typing a flag for non required flags complete. Which kinda makes sense. I think maybe the alternative would be more annoying: it would always add a '-' when you tab because there surely there are always more flags hehe

aslakknutsen:
This is more of a theoretical question:

    When the current folder contains a .ike.config.yaml or '-c' is used, e.g. -d might technically not be 'required' anymore

e2e/e2e_suite_test.go
```diff
@@ -38,53 +38,74 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
+		deployHelloWorldCmd("my-datawire-deployment"),
+		"oc new-project otherproject",
+		deployHelloWorldCmd("other-1-datawire-deployment"),
+		deployHelloWorldCmd("other-2-datawire-deployment"),
+		"oc project myproject",
+	)
+}
+
+func deployHelloWorldCmd(name string) string {
+	return "oc new-app --docker-image datawire/hello-world --name "+name+" --allow-missing-images"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	return "oc new-app --docker-image datawire/hello-world --name " + name + " --allow-missing-images"
```

PR #185, opened 2019-07-10
fix(macos): ensures istio-workspace builds on MacOS

### LINKED ISSUE #184: MacOS build is failing
##### Issue Overview

Building the project on MacOS does not work :(

##### Expected Behaviour

`make` passes

##### Current Behaviour

Multiple failures observed

- `operator-sdk` in always a linux binary which failes codegen under MacOS
- test failures
  - tmp directory path contains extra `/` - see https://github.com/golang/go/issues/21318
  - watch cmd test results in goroutine leak
- `ike` binary is always compiled with `GOOS=linux` resulting in non working binary under MacOS
- `ike` binary for docker build should still be build targeting `linux` as the container is based on `centos:7`

##### Steps To Reproduce

1. Clone project on MacOS
2. `make tools` and `make` afterward

bartoszmajsak:

It now works on MacOS \o/

- fix: downloads os-specific operator-sdk
- fix: builds os-specific binaries (and linux for docker)
- fix: proper tmp dir for macOS

Fixes #184

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,14 @@ PACKAGE_NAME:=github.com/maistra/istio-workspace
 OPERATOR_NAMESPACE?=istio-workspace-operator
 TEST_NAMESPACE?=bookinfo
 
+ifeq ($(shell uname), Linux)
+PLATFORM = linux-gnu
+else
+PLATFORM = apple-darwin
+endif
+
+ARCH:=$(shell uname -m)
+
 CUR_DIR:=$(shell pwd)
 export CUR_DIR
 BUILD_DIR:=$(CUR_DIR)/build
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -4,6 +4,14 @@ PACKAGE_NAME:=github.com/maistra/istio-workspace
 OPERATOR_NAMESPACE?=istio-workspace-operator
 TEST_NAMESPACE?=bookinfo
 
+ifeq ($(shell uname), Linux)
+PLATFORM = linux-gnu
+else
+PLATFORM = apple-darwin
```
bartoszmajsak: This is for `operator-sdk` - wondering if deserves a comment. WDYT?

test/tmp_file.go
```diff
@@ -29,7 +30,8 @@ var appFs = afero.NewOsFs()
 func TmpDir(t TestReporter, dir string) string {
 	fullPath := dir
 	if !path.IsAbs(dir) {
-		fullPath = fmt.Sprintf("%s/%s/%s", os.TempDir(), randomAlphaNumeric(), dir)
+		tmpDir := filepath.Clean(os.TempDir())
```
bartoszmajsak: Fixes https://github.com/golang/go/issues/21318 - need a comment?

aslakknutsen: A quick comment would be helpful. Else it's always in history but...

Makefile
aslakknutsen: PLATFORM is hardcoded in a similar way in the operators-sdk scripts?

aslakknutsen: any reason to align our .goreleaser.yml naming?

bartoszmajsak: The `PLATFORM` here is for `operator-sdk` download only. Maybe it deserves a better name or even being defined in the target itself.

> PLATFORM is hardcoded in a similar way in the operators-sdk scripts?

I am not sure what you mean by scripts here.

> any reason to align our .goreleaser.yml naming?

No :) I can understand the reasoning that it's easier to understand `apple-darwin` than just `Darwin`, but it makes downloading/automation harder because it doesn't match output from `uname`.

```
$ uname -s
Linux

$ uname -p
x86_64
```

This way your download link simply becomes `ike_${VERSION}_$(uname -s)_$(uname -p).tar.gz`

aslakknutsen: operator-sdk scripts=their release process. As to why/how they possible ended up with that naming. :)

bartoszmajsak: Not sure, but I will investigate.

investigation-fbi

test/tmp_file.go
bartoszmajsak: e2c3fbd

Makefile
bartoszmajsak: 18525bf

PR #187, opened 2019-07-13
chore(docker): update to rh ubi8 minimal base image

aslakknutsen:


bartoszmajsak:
Interesting. I have a branch with that open for a while and never managed to get the e2e build running https://github.com/bartoszmajsak/istio-workspace/tree/ubi_image

build/DockerfileTest
```diff
@@ -1,7 +1,4 @@
-FROM centos:7
-
-RUN adduser istio-workspace
-USER istio-workspace
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
bartoszmajsak: Are we ok with moving target as a base image?

aslakknutsen: Well.. Not really, but it's technically not any more moving than the one we had, as it's version locked by name, not tag. centos:7 vs uib8:latest would be same same was my thinking.

bartoszmajsak: Can we lock it to latest `8.0-131` then?

aslakknutsen: 8.0?

bartoszmajsak: yeah `8.0` would do, even though == latest :)

bartoszmajsak:
```suggestion
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.0
```

build/Dockerfile
```diff
@@ -1,6 +1,3 @@
-FROM centos:7
-
-RUN adduser istio-workspace
-USER istio-workspace
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
bartoszmajsak:
```suggestion
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.0
```

PR #198, opened 2019-07-23
fix(api): apply correct json tag to inline resource

aslakknutsen:
ops


bartoszmajsak:
Makes me wonder why `golangci` missed that one... not to mention tests :>

PR #200, opened 2019-07-24
feat(scan): scan for more virtual services to mutate

### LINKED ISSUE #138: Locate all references to a host/virtual service to setup the new route
Currently, we only support to manipulate the direct virtual service that is named as the "dc derived host".

As a first step, we should locate all virtual services that contain a reference to the "dc derived host".

E.g. a Gateway connected VirtualService that refer to our target dc "productpage"

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: bookinfo
spec:
  hosts:
  - "*"
  gateways:
  - bookinfo-gateway
  http:
  - match:
    - uri:
        exact: /productpage
    - uri:
        exact: /login
    - uri:
        exact: /logout
    - uri:
        prefix: /api/v1/products
    route:
    - destination:
        host: productpage
        port:
          number: 9080
```
'match' need to be cloned and restricted with the given header set and
'destination' needs to be updated to point to the manipulated subset

aslakknutsen:
fixes #138

### DIFF
--- a/pkg/istio/virtualservice.go
+++ b/pkg/istio/virtualservice.go
@@ -37,8 +37,8 @@ func VirtualServiceMutator(ctx model.SessionContext, ref *model.Ref) error { //n
 		return err
 	}
 
-	for _, vs := range vss.Items {
-		if !mutationRequired(&vs, targetHost, targetVersion) {
+	for _, vs := range vss.Items { //nolint[:rangeValCopy]
+		if !mutationRequired(vs, targetHost, targetVersion) {
 			continue
 		}
 		ctx.Log.Info("Found VirtualService", "name", targetHost)
@@ -189,7 +189,7 @@ func getVirtualServices(ctx model.SessionContext, namespace string) (*istionetwo
 	return &virtualServices, err
 }
 
-func mutationRequired(vs *istionetwork.VirtualService, targetHost, targetVersion string) bool {
+func mutationRequired(vs istionetwork.VirtualService, targetHost, targetVersion string) bool { //nolint[:hugeParam]
 	for _, http := range vs.Spec.Http {
 		for _, route := range http.Route {
 			if route.Destination != nil && route.Destination.Host == targetHost {
--- a/pkg/istio/virtualservice.go
+++ b/pkg/istio/virtualservice.go
@@ -7,6 +7,7 @@ import (
 	"github.com/maistra/istio-workspace/pkg/model"
 
 	istionetwork "istio.io/api/pkg/kube/apis/networking/v1alpha3"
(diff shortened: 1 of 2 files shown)


pkg/istio/virtualservice.go
```diff
@@ -26,28 +27,34 @@ func VirtualServiceMutator(ctx model.SessionContext, ref *model.Ref) error { //n
...
 		return err
 	}
 
-	ctx.Log.Info("Found VirtualService", "name", targetName)
-	mutatedVs, err := mutateVirtualService(ctx, ref.Target.GetNewVersion(ctx.Name), *vs)
-	if err != nil {
-		ref.AddResourceStatus(model.ResourceStatus{Kind: VirtualServiceKind, Name: targetName, Action: model.ActionFailed})
-		return err
-	}
+	for _, vs := range vss.Items {
```
golangcibot: rangeValCopy: each iteration copies 388 bytes (consider pointers or indexing) (from `gocritic`)

pkg/istio/virtualservice.go
```diff
@@ -26,28 +27,34 @@ func VirtualServiceMutator(ctx model.SessionContext, ref *model.Ref) error { //n
...
+		if !mutationRequired(&vs, targetHost, targetVersion) {
```
golangcibot: Using a reference for the variable on range scope `vs` (from `scopelint`)

bartoszmajsak:
@aslakknutsen what keeps this PR in draft? Can you highlight missing bits so it would be easier for me to get the full picture?

aslakknutsen:
Ah yeah, should have noted that.

I believe it will clash when https://github.com/Maistra/istio-workspace/pull/154 is merged. You can review the content of this, but it needs a few adjustments when the other is merged. This was set to draft just to ensure order :)

pkg/istio/virtualservice_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	"istio.io/api/networking/v1alpha3"
```
golangcibot: other declaration of v1alpha3 (from `typecheck`)

pkg/istio/virtualservice_test.go
```diff
@@ -194,6 +195,51 @@ var _ = Describe("Operations for istio VirtualService kind", func() {
 			})
 		})
 	})
+
+	Context("required", func() {
+		BeforeEach(func() {
+			virtualService = istionetwork.VirtualService{
```
golangcibot: undeclared name: `virtualService` (from `typecheck`)

PR #207, opened 2019-08-01
feat: updates project to be used against Maistra 0.12

### LINKED ISSUE #196: Update e2e-tests to Maistra 0.12

bartoszmajsak:

Brings both cluster and the code up to date with Maistra 0.12

There are several changes coming with this PR

##### E2E Setup

- There is no `istiooc` binary anymore
  - New scripts are provided to configure and start the cluster and then enable Maistra on it
- Cluster starts before the test suite as a separated step in circleci
  - Fixed logic in the test suite on how skipping cluster startdown and shutdown is handled
  - Now you can specify through env vars cluster behaviour during the tests
    - `IKE_CLUSTER_DIR` defines the location where cluster config and volumes are kept 
    - `IKE_E2E_KEEP_NS` when set to `true` will keep namespaces/project after test execution (defaults to `false`) - this obviously might make subsequent tests failing if we run a cluster in a constrained environment such as CircleCI
    - `IKE_E2E_SKIP_CLUSTER_SHUTDOWN` will keep the cluster running after test suite execution
- Implemented (fixed) the logic of waiting for pod readiness

##### Our codebase

- New `apiGroups` has been added in the role so that our operator is able to communicate with Istio APIs

Fixes #196


deploy/istio-workspace/role.yaml
```diff
@@ -50,3 +50,9 @@ rules:
       - '*'
     verbs:
       - '*'
+  - apiGroups:
+      - networking.istio.io
```
bartoszmajsak: @aslakknutsen maybe we should align our `apiGroup` now to `sessions.istio.io`. WDYT?

scripts/openshift/deploy-istio.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+MAISTRA_VERSION=${MAISTRA_VERSION:-"0.12"}
+
+if [ -z "$1" ]; then
+    echo "-- Using default maistra version ${MAISTRA_VERSION}. You can override it by passing it as a second argument."
```
bartoszmajsak:
```
    echo "-- Using default maistra version ${MAISTRA_VERSION}. You can override it by passing it as a first argument."
```

deploy/istio-workspace/role.yaml
aslakknutsen: Technically we're not part of the istio project. Wonder if we should change it to something like maistra.io instead?

e2e/e2e_suite_test.go
```diff
@@ -33,17 +34,24 @@ var tmpPath = NewTmpPath()
 var _ = SynchronizedBeforeSuite(func() []byte {
 	tmpPath.SetPath(path.Dir(shell.CurrentDir())+"/dist", os.Getenv("PATH"))
 	ensureRequiredBinaries()
+
+	if skip, found := os.LookupEnv("IKE_SKIP_CLUSTER_SHUTDOWN"); found {
```
aslakknutsen: Should we align the names of these vars, e.g. all are IKE_E2E prefixed?

scripts/dev/deploy-operator.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
aslakknutsen: This script is a pure 'dev' thing, correct? Doesn't seem to be used by e2e

deploy/istio-workspace/role.yaml
bartoszmajsak: I meant `maistra`. That was pre-coffee

scripts/dev/deploy-operator.sh
bartoszmajsak: Yeah, that's for my laziness. And that's the reason why its in `dev` folder :)

scripts/dev/deploy-operator.sh
```diff
@@ -0,0 +1,15 @@
...
+
+export IKE_DOCKER_REGISTRY=docker-registry-default.127.0.0.1.nip.io:80
+export IKE_DOCKER_REPOSITORY=istio-workspace-operator
+export TELEPRESENCE_VERSION=$(telepresence --version)
+export IKE_IMAGE_TAG="$(make version)"
+
+oc new-project istio-workspace-operator || true
+oc adm policy add-scc-to-user anyuid -z default -n $(oc project -q)
```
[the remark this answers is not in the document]
bartoszmajsak: So you are saying we shall remove it?

e2e/e2e_suite_test.go
bartoszmajsak:
```suggestion
	if skip, found := os.LookupEnv("IKE_E2E_SKIP_CLUSTER_SHUTDOWN"); found {
```

e2e/e2e_suite_test.go
```diff
@@ -33,17 +34,24 @@ var tmpPath = NewTmpPath()
...
-				"--enable", "registry,router,persistent-volumes,istio,centos-imagestreams",
-				"--base-dir", tmpClusterDir+"/maistra.local.cluster",
-			).Done()
+			projectDir := os.Getenv("PROJECT_DIR")
+			Expect(os.Setenv("IKE_CLUSTER_DIR", tmpClusterDir)).ToNot(HaveOccurred())
+			<-testshell.ExecuteInDir(projectDir, "make", "start-cluster").Done()
 		})
-		skipClusterShutdown = true
+	} else {
+		if _, found := os.LookupEnv("IKE_SKIP_CLUSTER_SHUTDOWN"); !found {
```
bartoszmajsak:
```suggestion
		if _, found := os.LookupEnv("IKE_E2E_SKIP_CLUSTER_SHUTDOWN"); !found {
```

scripts/dev/deploy-operator.sh
aslakknutsen: Ah! didn't notice the folder :)

scripts/dev/deploy-operator.sh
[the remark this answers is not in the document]
aslakknutsen: yea

deploy/istio-workspace/role.yaml
aslakknutsen: https://github.com/Maistra/istio-workspace/issues/208

e2e/e2e_suite_test.go
```diff
@@ -49,7 +49,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 			<-testshell.ExecuteInDir(projectDir, "make", "start-cluster").Done()
 		})
 	} else {
-		if _, found := os.LookupEnv("IKE_SKIP_CLUSTER_SHUTDOWN"); !found {
+		if _, found := os.LookupEnv("IKE_E2E_SKIP_CLUSTER_SHUTDOWN"); !found {
```
aslakknutsen: The CLUSTER_DIR too?

bartoszmajsak: Was thinking about that one, but I am not entirely sure, as the script can be used not only by e2e tests.

aslakknutsen: ok.

PR #211, opened 2019-08-02
feat: adds opencontainers labels to operator image

### LINKED ISSUE #178: Introduce labels for the images
##### What can be improved

https://github.com/opencontainers/image-spec/blob/master/annotations.md

##### Why

>Many tools need some information about a container image, or about the consumer’s use of that container image. Metadata built into container images can enable even more powerful container management tooling.
> 
>By agreeing on a conventional set of labels, everyone can use tools from different vendors without the need to add vendor-specific labels to all their images, especially where those labels carry the same information or serve the the same purpose. Duplication leads to the risk of error.

bartoszmajsak:
Fixes #178


bartoszmajsak:
Not sure if we have any k8s metadata in mind which we could add to deployment yaml. WDYT @aslakknutsen?

aslakknutsen:
@bartoszmajsak Maybe add labels to the test-service image as well?

aslakknutsen:
@bartoszmajsak Not sure about k8 metadata. Suspect some will come with the Operator Lifecycle thinggy

bartoszmajsak:
> @bartoszmajsak Maybe add labels to the test-service image as well?

I will with the new PR - coming soon.

Issue #214, opened 2019-08-08
Add support for multiple operators running on the same cluster


aslakknutsen:
##### What can be improved

The operator should be able to only react to events in namespaces a given operator should operate. 

A simple solution would be:

Annotate the namespace with e.g. maistra.io/istio-workspace-operator: "operator-namespace"
Then only the operator in "operator-namespace" would react to those events.

##### Why

When running CI jobs or multiple operator developers working on the same cluster we require the operators to not all react to the same namespaces.

bartoszmajsak:
How should handle default case - so preserve the current functionality? If ns has no annotation but we have multiple instances of operator running - what should be the behaviour?

aslakknutsen:
That is a very good question hehe :)

We might add an Env Var on the Deployment or something to set it in 'filtered' mode, then only react to the ones that match via annotation. So you have 1 that is cluster installed that ignore namespaces with the annotation, then the others that only react to the correct annotated namespace... ?

bartoszmajsak:
OTOH this is a special case which should always be configured - so maybe `ike` cli should take care of propagating the annotation? so if nothing specified we set it to `default` and we always have an operator which will handle that. other than that you might be able to query for possible values when `--target-operator` or however we name the annotation?

aslakknutsen:
That would move the 'config' to the Session object....

How would you resolve the targets?

bartoszmajsak:
Not sure if I understand your concern. My thinking is - at the moment we target the deployment we know where it lives, so we could annotate that ns (if it's not already) while establishing "connection".

PR #215, opened 2019-08-09
feat(dev): deploy operator to single namespace

aslakknutsen:
Add make target (un)deploy-operator-local to bind the operator to a single namespace.

OPERATOR_NAMESPACE=bookinfo make deploy-operator-local

Related #214


aslakknutsen:
This generally works, but a few problems:

* only 'local' Operators can be installed on a single cluster, not combined with a Custer wide
* the deploy/undeploy install/uninstall the CRD as well so technically 1 can mess up for the others...

Makefile
```diff
@@ -239,7 +240,8 @@ define process_template # params: template location
 		-p IKE_IMAGE_NAME=$(IKE_IMAGE_NAME) \
 		-p IKE_IMAGE_TAG=$(IKE_IMAGE_TAG) \
 		-p NAMESPACE=$(OPERATOR_NAMESPACE) \
-		-p TELEPRESENCE_VERSION=$(TELEPRESENCE_VERSION)
+		-p TELEPRESENCE_VERSION=$(TELEPRESENCE_VERSION) \
+    -p WATCH_NAMESPACE=$(OPERATOR_WATCH_NAMESPACE)
```
bartoszmajsak:
```suggestion
		-p WATCH_NAMESPACE=$(OPERATOR_WATCH_NAMESPACE)
```

Makefile
```diff
@@ -262,13 +264,35 @@ deploy-operator: ## Deploys istio-workspace operator resources to defined OPERAT
...
 	$(call process_template,deploy/istio-workspace/role_binding.yaml) | oc delete -n $(OPERATOR_NAMESPACE) -f -
 	oc delete -n $(OPERATOR_NAMESPACE) -f deploy/istio-workspace/role.yaml
 	oc delete -n $(OPERATOR_NAMESPACE) -f deploy/istio-workspace/service_account.yaml
 	oc delete -n $(OPERATOR_NAMESPACE) -f deploy/istio-workspace/crds/istio_v1alpha1_session_crd.yaml
 
+.PHONY: deploy-operator-local
+deploy-operator-local: export OPERATOR_WATCH_NAMESPACE=$(OPERATOR_NAMESPACE)
+deploy-operator-local: ## Deploys istio-workspace operator resources to a single Namespace defined by OPERATOR_NAMESPACE
+	$(call header,"Deploying local operator to $(OPERATOR_NAMESPACE)")
+	export OPERATOR_WATCH_NAMESPACE=$(OPERATOR_NAMESPACE)
```
bartoszmajsak: Any reason why we have this here and in `275`?

deploy/istio-workspace/operator.yaml
```diff
@@ -17,6 +17,10 @@ parameters:
     description: "The tag of the image to be used"
     required: true
     value: latest
+  - name: WATCH_NAMESPACE
+    description: "The tag of the image to be used"
```
bartoszmajsak: nene

aslakknutsen: :D https://github.com/Maistra/istio-workspace/pull/215/commits/27fdf4fbbafa5534a5f30a05aeda390197b04209

Makefile
aslakknutsen: Oh, there are reasons... but none that are valid. https://github.com/Maistra/istio-workspace/pull/215/commits/649f78fbd08f09a429d47264d544de66ee8d4bce

aslakknutsen:
> Did you try to use it with ike install-operator -n local_ns?

Nope, not yet. Was thinking of maybe adding a --local flag to the install-operator ?

> Should we think about having a e2e test for it (issue here is that we have global operator installed beforehand)?

I was thinking more we change the e2e to use this on PR runs

aslakknutsen:
> Why is your build failing?

Because you've hardcoded the whole Deployment in a test ;)

bartoszmajsak:
> > Why is your build failing?
> 
> Because you've hardcoded the whole Deployment in a test ;)

It's a GOLDEN DEPLOYMENT

PR #216, opened 2019-08-09
feat: adds labels to operator deployment

bartoszmajsak:

Adds `app` and `version` label to the operator deployment


pkg/template/template_test.go
```diff
@@ -171,7 +171,7 @@ var _ = Describe("Operations for template system", func() {
 				Expect(string(o)).To(ContainSubstring("DEFAULT_VERSION"))
 			})
 
-			It("should override with incomming is available", func() {
+			It("should override with incoming is available", func() {
```
aslakknutsen:
```suggestion
			It("should override with incoming if available", func() {
```

PR #219, opened 2019-08-12
chore(scripts): add kiali and jeager to deploy scripts

aslakknutsen:


aslakknutsen:
Hmm.. this is currently used by CI... which will to some extent be removed? Will this be moved to dev or? @bartoszmajsak

bartoszmajsak:
Let's leave it unchanged as this is only for CI and we have resource constraints there. Plus it's for non-RHEL8 installation of OKD, such as `oc cluster up` with OKD 3.11 we use on the Circle server (it's some version of Ubuntu).

Maybe it should not sit in the current folder, as it suggests its usage is wider than that. 

Let's have another one for dev purposes in the dedicated folder and this one we probably should move to `.circleci` folder. WDYT?

aslakknutsen:
That sounds reasonable.

aslakknutsen:
not required anymore...

PR #220, opened 2019-08-12
fix(develop): allow to join / leave session

aslakknutsen:
fixes failure on Creating a resource with a resourceVersion set.


aslakknutsen:
@bartoszmajsak hehe I did remember you mention the error earlier, but couldn't see a PR so.. :)

PR #221, opened 2019-08-14
build(deps): bumps golang to 1.12.8 patch release

bartoszmajsak:


bartoszmajsak:
It seems that status on a PR is in limbo, as both builds were successfully executed for the commit.

PR #222, opened 2019-08-20
Testing Actions

aslakknutsen:


bartoszmajsak:
Are we planning to use that workflow? Otherwise let's close that PR. Seems to be in a hiatus right now :)

PR #224, opened 2019-08-21
feat(tests): adjusts e2e tests to be ran against 4.1 clusters

bartoszmajsak:

There are some differences between Openshift 3.11 (currently used) and 4.1. This PR addressed them.

You can run this test suite against any cluster now using:

```
IKE_E2E_MANAGE_CLUSTER=false IKE_CLUSTER_USER=admin-user IKE_CLUSTER_PWD=admin-IKE_CLUSTER_HOST=xxx.yyy.zzz ISTIO_NS=istio-system-ns make test-e2e
```

ISTIO_NS is not needed and defaults to `istio-system`, but in case of our QE cluster is should be set to `service-mesh-1`

- allows specifying admin user credential through env var
- starts istio only when the cluster is managed by e2e tests (defined through env var)
- checking cluster version to determine how to enable required services (as they changed their names)
- parsing shell scripts using shell lexer (so less error-prone then before -> less wrapping in `bash -c`)
- shuffled helper funcs around


bartoszmajsak:
@aslakknutsen I can configure circleci to have extra job running against 4.1 cluster as part of this PR. WDYT?

aslakknutsen:
@bartoszmajsak Yeah sounds good. We end up with 3.11 local and 4.1 remote right? (jog my memory :)

bartoszmajsak:
> @bartoszmajsak Yeah sounds good. We end up with 3.11 local and 4.1 remote right? (jog my memory :)

That's correct.

bartoszmajsak:
Closing this PR so that I can push it to the upstream branch and re-open.

Issue #61, opened 2019-03-04
API for choosing clone strategy


aslakknutsen:
* Teleprecence
* Pure deployment
* Squash?

```yaml
spec:
  refs:
    - name: x-deploy
      strategy: telepresence
      args:
        version: 100.2
    - name: y-deploy
      strategy: image
      args:
        image: x/org:latest
```

aslakknutsen:
fixed by https://github.com/Maistra/istio-workspace/pull/235

PR #235, opened 2019-08-29
Ref strategy

aslakknutsen:
Related to #61


pkg/internal/session/session_client_test.go
```diff
@@ -1,6 +1,7 @@
 package session_test
 
 import (
+	"github.com/maistra/istio-workspace/pkg/apis/istio/v1alpha1"
```
golangcibot: dupImport: package is imported 2 times under different aliases on lines 4 and 5 (from `gocritic`)

PR #250, opened 2019-09-04
feat: dumps logs for failing e2e tests

### LINKED ISSUE #175: Show logs of failed pods
##### What can be improved

When the build fails because the pod is in `RunContainerError` for example we won't know the exact cause unless we re-run the suite and grab the logs. This is far from optimal. We should dump logs and events of failing pods.

##### Why

Easier troubleshooting of our test-infra.

bartoszmajsak:

Improves diagnostic of e2e test failures.

- dumps logs of all pods in the test namespace
- dumps the state of each container in the pod
- dumps events

Fixes #175

### DIFF
--- a/e2e/infra/pod_state.go
+++ b/e2e/infra/pod_state.go
@@ -10,18 +10,13 @@ import (
 // AllPodsReady checks whether all the pods (and their containers) in the given namespace are in Ready state
 func AllPodsReady(ns string) func() bool {
 	return func() bool {
-		podsCmd := shell.ExecuteInDir(".",
-			"oc", "get", "pod",
-			"-n", ns,
-			"-o", "jsonpath={.items[*].metadata.name}")
-		<-podsCmd.Done()
-		pods := strings.Split(strings.Trim(fmt.Sprintf("%s", podsCmd.Status().Stdout), "[]"), " ")
+		pods := GetAllPods(ns)
 		for _, pod := range pods {
 			if strings.Contains(pod, "-deploy") {
 				fmt.Printf("Skipping deploy pod %s\n", pod)
 				continue
 			}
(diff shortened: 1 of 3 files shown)


e2e/infra/pod_state.go
```diff
@@ -30,7 +25,30 @@ func AllPodsReady(ns string) func() bool {
...
+		"oc", "get", "pod",
+		"-n", ns,
+		"-o", "jsonpath={.items[*].metadata.name}")
+	<-podsCmd.Done()
+	return strings.Split(strings.Trim(fmt.Sprintf("%s", podsCmd.Status().Stdout), "[]"), " ")
+}
+
+// StateOf returns state of the pod
+func StateOf(ns, pod string) {
+	state := shell.Execute("oc get pod " + pod + " -n " + ns + " -o yaml")
```
aslakknutsen: Did we conclude on the oc vs kubectrl?

bartoszmajsak: I was thinking about this one, but I am not sure how to break it down. This is not only about using `kubectl` but also having tests ran against vanilla k8s with (presumably) vanilla istio. This would require not only changing CLI tools invocations but also assumptions about the cluster (e.g. we do create some maistra-specific CR right now).

We could have one task switching from `oc` to `kubectl` at first and then the `k8s` one as separated step, but I am not sure if split would bring any value. WDYT?

bartoszmajsak: If our goal is to have `ike` working with any `k8s` distro then we should probably go for it as preps on what else should we change in the production code too.

aslakknutsen: We can start by writing down some basics in an Epic or something and split out tasks from there.

PR #251, opened 2019-09-05
feat(cli): prints version hint when outdated

### LINKED ISSUE #14: Version update hint
While running CLI check if there's newer version released and suggest user the update path

bartoszmajsak:

Whenever `ike` cli gets new release it will prompt the user with a warning.

```
╰─ ike develop                         
WARN: you are using v0.0.1 which is not the latest release (newest is v0.0.2). You can update it using:
$ curl -sL http://git.io/get-ike | bash
```

- ability to indicate if binary is an official release through vars when building
- uses gh api to determines if it is latest release
- prints warning about outdated version for every command

Fixes #14


pkg/cmd/root.go
```diff
@@ -27,6 +29,11 @@ func NewCmd() *cobra.Command {
 		Short:                  "ike lets you safely develop and test on prod without a sweat",
 		BashCompletionFunction: completion.BashCompletionFunc,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error { //nolint[:unparam]
+			if v.Released() && !version.IsLatestRelease() {
```
aslakknutsen: Calling GitHub twice, IsLatestRelease and LatestRelease

Maybe instead:

latest := LatestRelease()
if version.IsLatest(latest) {}

aslakknutsen: On missing network, this would potentially timeout twice

aslakknutsen: This being in the root command.. that would force 'serve' to check for version on startup too?

aslakknutsen: Could we do it 'async' in Pre and show the message in Post or something to avoid it interfering with normal operations?

pkg/cmd/root.go
```diff
@@ -27,6 +29,11 @@ func NewCmd() *cobra.Command {
...
+				latestRelease, _ := version.LatestRelease()
+				fmt.Printf("WARN: you are using %s which is not the latest release (newest is %s). "+
+					"You can update it using:\n$ curl -sL http://git.io/get-ike | bash\n", v.Version, latestRelease)
```
aslakknutsen: Technically the suggested script could be wrong, couldn't it. It assume ike is in the current dir? 

    curl .../get-ike |  bash -s — --version=v0.0.1 --dir=~`type ike*magic-parent-dir`

pkg/cmd/root.go
bartoszmajsak: > This being in the root command.. that would force 'serve' to check for version on startup too?

Yup, is that a concern? It won't fail, will log in case of network issues.

bartoszmajsak: > Could we do it 'async' in Pre and show the message in Post or something to avoid it interfering with normal operations?

I will have a look. Thanks for the hint.

bartoszmajsak: > Maybe instead:
> 
> latest := LatestRelease()
> if version.IsLatest(latest) {}

Good point.

pkg/cmd/root.go
bartoszmajsak: Then maybe we should simply link to docs and fix that there?

pkg/cmd/root.go
bartoszmajsak: Single gh call fixed in 694dfd8

aslakknutsen: > > Could we do it 'async' in Pre and show the message in Post or something to avoid it interfering with normal operations?
> 
> I will have a look. Thanks for the hint.

Or maybe even simpler have a flag to turn it off. On by default but can be turned off.

aslakknutsen: > Yup, is that a concern? It won't fail, will log in case of network issues.

The concern is more around it wouldn't be the first 'un-required' thing in a program to fail or slow it down to a dog. Good intentions and all but.. Let's ensure it doesn't get in the way.

pkg/cmd/root.go
aslakknutsen: That would be a safe bet. Maybe link to release notes and include install notes there. In case some migration type of thing would be needed as well?

pkg/cmd/root.go
bartoszmajsak: Reworked in 9369c11

pkg/cmd/root.go
bartoszmajsak: b23fb82

pkg/cmd/root.go
```diff
@@ -21,12 +24,29 @@ var log = logf.Log.WithName("cmd").WithValues("type", "root")
...
+	timeout := make(chan bool, 1)
 
 	rootCmd := &cobra.Command{
 		Use:                    "ike",
 		Short:                  "ike lets you safely develop and test on prod without a sweat",
 		BashCompletionFunction: completion.BashCompletionFunc,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error { //nolint[:unparam]
+			if v.Released() {
+				go func() {
+					time.Sleep(2 * time.Second)
```
aslakknutsen: time.NewTimer ?

bartoszmajsak: Thanks, that makes it easier! fd4a3bd

PR #257, opened 2019-09-12
fix: handles error coming from root cmd

bartoszmajsak:
I guess we shouldn't swallow error thrown from `root` command :) Seems like left-over from some initial spike.


bartoszmajsak:
Closing as PR build against the remote cluster is failing permanently.

Reason: PR was open from forked repo - thus env vars in CirleCI are not propagated.

PR #260, opened 2019-09-13
fix: makes padding bigger

bartoszmajsak:
With new target names being quite excessive we didn't have target - explanation nicely aligned. This PR fixes that.


bartoszmajsak:
It's l0ExwJPrLcCGfXHck

after all!

PR #272, opened 2019-09-23
feat: supports global envs for config

### LINKED ISSUE #271: Allow global defined env variables
##### What can be improved

Define configuration/env variables on an 'ike' level and not just per command to avoid repeating the same configuration.

##### Why

Since we now have multiple cli commands that rely on the same parameters it could be useful to allow to define them on a global scale and not just per command. E.g. Namespace, Deployment, Session

Instead of:

```
IKE_CREATE_NAMESPACE=x
IKE_CREATE_DEPLOYMENT=y
IKE_CREATE_SESSION=z
IKE_DELETE_NAMESPACE=x
IKE_DELETE_DEPLOYMENT=y
IKE_DELETE_SESSION=z
```

Something like:

```yaml
namespace: x
deployment: ratings-v1
develop:
  session: x
create:
  session: y
delete:
  session: y
 ```

```
IKE_NAMESPACE=x
IKE_DEPLOYMENT=ratings-v1
IKE_SESSION=test
IKE_CREATE_ROUTE=y
IKE_CREATE_IMAGE=y
```

aslakknutsen:
fixes #271

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -503,7 +503,7 @@
   version = "v0.7.0"
 
 [[projects]]
-  digest = "1:b211368aa39cabd94f594cdbe7330ce6f7df495e68b9c3d4fecb457772616240"
+  digest = "1:55bcf4202ec2033988abce4932b6ad0f424fbd3d408e80368c24a23c406998d8"
   name = "github.com/prometheus/procfs"
   packages = [
     ".",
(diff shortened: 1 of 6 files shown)


bartoszmajsak:
LGTM so far. What's missing?

aslakknutsen:
@bartoszmajsak I was thinking some tests.. hehe

bartoszmajsak:
> @bartoszmajsak I was thinking some tests.. hehe

that was obvious so didn't even bother to ask ;)

PR #275, opened 2019-10-01
fix: enables use of IKE_CONFIG env var to pass config file

### LINKED ISSUE #274: Unable to specify config file through env variable
##### Issue Overview

The only way to define different config file than default `.ike.config.yaml` is by `-c|--config` flag. It should be however possible to overwrite its location by using `IKE_CONFIG` env variable.

##### Expected Behaviour

`IKE_CONFIG` should take precedence over.

##### Current Behaviour

Not possible to use `IKE_CONFIG`.

bartoszmajsak:
Fixes #274

### DIFF
--- a/e2e/smoke_test.go
+++ b/e2e/smoke_test.go
@@ -99,92 +99,94 @@ var _ = Describe("Smoke End To End Tests - against OpenShift Cluster with Istio 
 			cleanupNamespace(namespace)
 		})
 
-		Context("scenario-1-basic-deployment", func() {
+		Describe("k8s deployment", func() {
+
 			BeforeEach(func() {
 				scenario = "scenario-1"
 			})
 
-			It("should watch for changes in ratings service and serve it", func() {
-				verifyThatResponseMatchesModifiedService(tmpDir, namespace)
+			Context("basic deployment modifications", func() {
+				It("should watch for changes in ratings service and serve it", func() {
+					verifyThatResponseMatchesModifiedService(tmpDir, namespace)
+				})
+
+				It("should watch for changes in ratings service in specified namespace and serve it", func() {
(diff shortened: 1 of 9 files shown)


pkg/cmd/config/config.go
```diff
@@ -27,13 +36,9 @@ import (
...
-func SetupConfigSources(configFile string, notDefault bool) error {
-	viper.Reset()
-	viper.SetEnvPrefix("IKE")
-	viper.AutomaticEnv()
-	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
-	viper.SetTypeByDefaultValue(true)
-
+func SetupConfigSources(configFile string, defaultConfigFile bool) error {
+	println(configFile)
+	println(defaultConfigFile)
```
aslakknutsen: ?

bartoszmajsak: LOL, suggestion would help here haha

bartoszmajsak: Fixed with 96ba756

e2e/smoke_test.go
```diff
@@ -191,7 +191,8 @@ var _ = Describe("Smoke End To End Tests - against OpenShift Cluster with Istio
 		// Thus it treats cluster as vanilla k8s and expects Deployment, not DeploymentConfig to appear
 		// Enable when https://github.com/telepresenceio/telepresence/issues/1139 is fixed
 		XContext("openshift deploymentconfig", func() {
-			BeforeEach(func() {
+
+      BeforeEach(func() {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
			BeforeEach(func() {
```

Issue #236, opened 2019-08-30
Create an image for ike client to run in 


aslakknutsen:

bartoszmajsak:
Could you remind me what is the main driver for that? Is it to have CI step based on that? (i.e. when integrating with tekton?)

aslakknutsen:
@bartoszmajsak Correct

aslakknutsen:
See WIP https://github.com/aslakknutsen/istio-workspace-action

PR #276, opened 2019-10-01
feat(cli): return non 0 exit code on failure

aslakknutsen:
related to #236


aslakknutsen:
@bartoszmajsak 42 was too obvious

PR #282, opened 2019-10-07
feat: cleans up ServiceMeshMemberRoll spec after test execution

bartoszmajsak:


bartoszmajsak:
Pushed to the wrong branch (fork) which is the reason why remote e2e tests are failing. Will re-open with the right setup.

PR #292, opened 2019-10-31
feat: aligns telepresence version used by client and cluster

### LINKED ISSUE #284: Ensure that Telepresence image matches the binary on $PATH
##### What can be improved

Right now tp images version is determined through `TELEPRESENCE_VERSION` or fallback to hardcoded version. As we have `tp` binary on path we can get its version, pass it as strategy variable and use that for image deployment.

##### Why

Otherwise, we have a hardcoded version we have to maintain and update once in a while.

bartoszmajsak:
### Short description of what this resolves:

Sends the telepresence version used by the client (i.e. `ike develop`) to the server so the proper version of the image is used. Otherwise, if there's a mismatch, the command will fail highlighting this problem. This happens in the current implementation when we have hardcoded default and we don't send `TELEPRESENCE_VERSION` environment variable in the target deployment.

- checks version of telepresence using binary telepresence version resolution works as follows now:
  - `TELEPRESENCE_VERSION` env var takes precedence
  - if telepresence binary is available on the path then it will call `telepresence --version`
  - otherwise, we return hardcoded default version (last resort)
- introduces telepresence wrapper exposing functionalities used across the
pkgs
- tests tp version passed from the client by not setting env var for 4.x cluster e2e tests on CircleCI

Fixes #284

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -155,7 +155,6 @@ jobs:
       - run:
           name: "Runs end-to-end tests"
           command: |
-            TELEPRESENCE_VERSION=$(telepresence --version) \
             IKE_CLUSTER_USER=$QE_IKE_CLUSTER_USER \
             IKE_CLUSTER_PWD=$QE_IKE_CLUSTER_PWD \
             IKE_CLUSTER_HOST=$QE_IKE_CLUSTER_HOST \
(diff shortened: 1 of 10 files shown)


bartoszmajsak:
Another idea to mitigate hardcoded default version (fallback) would be to check latest GH release. But this would not prevent from still having older version on the client side. In general, I think we don't really need that hardcoded fallback at all right now. WDYT?

aslakknutsen:
Agree. The fallback won't really make anything work any better.

PR #294, opened 2019-11-01
feat: introduces env var handling for tests

bartoszmajsak:

Simplifies overwriting and restoring environment variables in tests.

Now you can simply do:

```golang
It("should process operator template", func() {
	// given
	envVars := test.TemporaryEnvVars()
	envVars.SetAll(map[string]string{
		"IKE_DOCKER_REGISTRY":   "quay.io",
		"IKE_DOCKER_REPOSITORY": "istio-workspace",
		"IKE_IMAGE_NAME":        "ike-cli",
		"IKE_IMAGE_TAG":         "latest",
	})
	defer func() {
		envVars.Restore()
	}()

	// ...
})
```

or by using ginkgo hooks:

```golang
envVars := TemporaryEnvVars()

BeforeEach(func() {
	envVars.Set("IKE_DEVELOP_PORT", "4321")
})

AfterEach(envVars.Restore)
```

### DIFF
--- a/pkg/cmd/develop/develop_test.go
+++ b/pkg/cmd/develop/develop_test.go
@@ -39,10 +39,10 @@ var _ = Describe("Usage of ike develop command", func() {
 		AfterEach(tmpPath.Restore)
 
 		It("should fail invoking develop cmd when telepresence binary is not on $PATH", func() {
-			oldPath := os.Getenv("PATH")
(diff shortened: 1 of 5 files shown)


aslakknutsen:
Why not return a 'close' function instead?

e.g. 

```go
restore := test.TempVars(map[string]string{"x":"y"})

restore()
```

PR #302, opened 2019-11-08
feat: adds ability to validate engine Patches

bartoszmajsak:

In order to ensure we always have `telepresence` version specified `Patch` can now have `Validate` func defined. In case of telepresence strategy template we simply ensure that `version` variable is passed before the template is processed.

- ability to define `func` for each engine `Patch`
- validates if `telepresence` strategy has `version` variable defined
- bunch of tests

Related to #292


pkg/template/template.go
```diff
@@ -55,6 +55,12 @@ func NewDefaultEngine() *Engine {
 			Variables: map[string]string{
 				"version": "",
 			},
+			Validate: func(vars map[string]string) error {
```
aslakknutsen: This becomes a little tricky when the templates are externalized(read from file) and any template could be used. Wondering if there is some kinda validation we could do in the Templates instead?

bartoszmajsak: Was pondering about that too. There's `isset` but I couldn't find if I can error from there.

aslakknutsen: Could make a custom function if none similar enough exist

bartoszmajsak: Done in 2243bd0

aslakknutsen: I think the If/Else option could be useful in some cases, but for the basic 'fail if missing' use case, couldn't we just have a single statement on top pr variable instead of the whole if/else construct, eg.

    {{ failIfVariableMissing "name" }}

?

bartoszmajsak: Was thinking about that but it will result in the weird function signature because for the templates the function has to return `(value, error)`. But to be honest the current `fail` is not much different :)

bartoszmajsak: Done in 73e4cc8

bartoszmajsak:
We need some docs for that feature. Sooner than later :)

Created issue #305 for that.

PR #308, opened 2019-11-14
fix(operator): mutate route if subset is missing

### LINKED ISSUE #307: VirtualService with no defined subset is not mutated
##### Issue Overview

A virtual service destination that is missing subset is not mutated.

##### Expected Behaviour

Any non-subset defined destination should be mutated and restricted to new v.

##### Current Behaviour

Ignored. The operator fails to find Route.

aslakknutsen:
If a mutation does not occur the original route will be hit and end up
round-robin between different versions.

fixes #307

### DIFF
--- a/pkg/istio/virtualservice.go
+++ b/pkg/istio/virtualservice.go
@@ -108,8 +108,8 @@ func mutateVirtualService(ctx model.SessionContext, sourceResource model.Located
 	}
 	removeOtherRoutes := func(http v1alpha3.HTTPRoute, host, subset string) v1alpha3.HTTPRoute {
 		for i, r := range http.Route {
-			if !((r.Destination.Host == host && r.Destination.Subset == subset) ||
-				(r.Destination.Host == host && r.Destination.Subset == "")) {
+			if !((r.Destination != nil && r.Destination.Host == host && r.Destination.Subset == subset) ||
+				(r.Destination != nil && r.Destination.Host == host && r.Destination.Subset == "")) {
 				http.Route = append(http.Route[:i], http.Route[i+1:]...)
 			}
 		}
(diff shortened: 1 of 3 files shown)


pkg/istio/virtualservice_test.go
```diff
@@ -152,6 +155,12 @@ var _ = Describe("Operations for istio VirtualService kind", func() {
 
 					Expect(GetMutatedRoute(mutatedVirtualService, targetV1Host, targetV1Subset).Redirect).To(BeNil())
 				})
+				It("include destinations with no subset", func() {
```
bartoszmajsak: I just wonder if it would bring some benefit to cover more boundary cases as the conditionals grew a bit. WDYT?

aslakknutsen: explain?

pkg/istio/virtualservice.go
```diff
@@ -97,16 +97,19 @@ func mutateVirtualService(ctx model.SessionContext, sourceResource model.Located
...
+					}
 				}
 			}
 		}
 		return routes
 	}
 	removeOtherRoutes := func(http v1alpha3.HTTPRoute, host, subset string) v1alpha3.HTTPRoute {
 		for i, r := range http.Route {
-			if !(r.Destination.Host == host && r.Destination.Subset == subset) {
+			if !((r.Destination.Host == host && r.Destination.Subset == subset) ||
```
bartoszmajsak: Could `r.Destination` be `nil` like in L100?

aslakknutsen: Technically it shouldn't be nil at this point. That's more of a copy paste error. The mutationRequired should block that. (if check to even enter this part). As well as Destination being required on the API level.

aslakknutsen: https://github.com/Maistra/istio-workspace/pull/308/commits/7f3de5bb207a9d796f96b2cdd97d307b0d1dcc4f

PR #312, opened 2019-11-28
fix(cmd): installs to current ns when --local mode used and no ns defined

### LINKED ISSUE #306: Local operator installation should take current namespace into consideration
##### Issue Overview

`ike install-operator -l|--local` installs operator to the local namespace. When `-n|--namespace` is not specified, however, it will default to `istio-workspace-operator` instead of the current active namespace (or project)

##### Expected Behaviour

Installs to the current namespace.

##### Current Behaviour

Installs to the `istio-workspace-operator` default namespace

##### Steps To Reproduce

1. Create new namespace|project
2. `ike install-operator -l` without specifying the namespace.

bartoszmajsak:

when `ike install-operator -l` is invoked it will install operator to the current namespace instead of `istio-workspace` default one which is intended for global.

- fix in the `install-operator` command
- e2e test covering local install path (I had quite a hard time figuring out how to test in more fine-grained level - e.g. using fake client or gock with stubbed calls. Those two paths seemed to be more expensive)

Fixes #306


e2e/operator_installation_test.go
```diff
@@ -0,0 +1,70 @@
...
+	Context("local installation", func() {
+
+		It("install ike operator into specified namespace", func() {
+			// given
+			projectName := "ike-local-installation-defined-namespace-" + naming.RandName(16)
+			<-testshell.Execute(NewProjectCmd(projectName)).Done()
+			defer func() {
+				<-testshell.Execute(DeleteProjectCmd(projectName)).Done()
+			}()
+			PushOperatorImage(projectName)
```
bartoszmajsak: I think it's time we should optimize it and have a shared image built as the previous step of the pipeline. WDYT?

e2e/smoke_test.go
```diff
@@ -78,7 +78,8 @@ var _ = Describe("Smoke End To End Tests - against OpenShift Cluster with Istio
 			<-testshell.Execute(NewProjectCmd(namespace)).Done()
 
 			UpdateSecurityConstraintsFor(namespace)
-			DeployLocalOperator(namespace)
+			PushOperatorImage(namespace)
+			InstallLocalOperator(namespace)
```
bartoszmajsak: Split into two steps as I need to push the image for the `install-operator` tests and only then invoke it in different flavours.

e2e/operator_installation_test.go
aslakknutsen:
```
build operator image
build test images

test
  tag/push operator image to ns
  tag/push test image to ns
```

?

e2e/operator_installation_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+			// when
+			<-testshell.Execute("ike install-operator -l -n " + projectName).Done()
```
aslakknutsen: call InstallLocalOperator(ns)?

e2e/operator_installation_test.go
```diff
@@ -0,0 +1,70 @@
...
+			// given
+			projectName := "ike-local-installation-defined-namespace-" + naming.RandName(16)
+			<-testshell.Execute(NewProjectCmd(projectName)).Done()
+			defer func() {
+				<-testshell.Execute(DeleteProjectCmd(projectName)).Done()
+			}()
+			PushOperatorImage(projectName)
+
+			// when
+			<-testshell.Execute("ike install-operator -l").Done()
```
aslakknutsen: call InstallLocalOperator(nil)?

e2e/operator_installation_test.go
bartoszmajsak: Sure, but isn't it hiding q bit the actual subject of the test?

e2e/operator_installation_test.go
bartoszmajsak: I was thinking about pushing to public repo so that we avoid test-level pushes too

aslakknutsen: Yeah, I was about to comment if we should just push to quay. I'm good with that.

pkg/cmd/install/install.go
```diff
@@ -48,11 +50,24 @@ func installOperator(cmd *cobra.Command, args []string) error { //nolint[:unpara
...
 	}
 
+	if local && !cmd.Flags().Changed("namespace") {
+		kubeCfg := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			clientcmd.NewDefaultClientConfigLoadingRules(),
+			&clientcmd.ConfigOverrides{},
+		)
+
+		if namespace, _, err = kubeCfg.Namespace(); err != nil {
+			log.Error(err, "failed to read k8s config")
```
aslakknutsen: Maybe make the error something like: Failed to read kubeconf 'config file path' to get the current namespace, please provide a namespace manually via -namespace?

e2e/operator_installation_test.go
bartoszmajsak: Created #313

e2e/operator_installation_test.go
```diff
@@ -0,0 +1,80 @@
+package e2e_test
+
+import (
+	"fmt"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	"fmt"

```

pkg/cmd/install/install.go
bartoszmajsak: Oh, I missed that one. Providing PR in a minute. #314

PR #342, opened 2020-01-09
fix(template): creates replica if not exists

### LINKED ISSUE #341: Template patching should check if path exists
##### Issue Overview

If `/spec/template/spec/replicas` does not exist in modified resource it fails with error (which is expected).

##### Expected Behaviour

We should first check the existence of the path and only then apply the patch.

##### Current Behaviour

Works ok, as jsonpatch library in version `v3.4.0` silently omits (?) the error. Upgrading to the newest version (transitively through open-ended versioning) yields `msg: "replace operation does not apply: doc is missing key: /spec/template/spec/replicas"`.

bartoszmajsak:

- check if `/spec/template/spec/replicas` exists. Creates if not before patching to `1`
- relaxes test assertion for wrong paths, as error msg changed between the versions of json-patch library

Fixes #341

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -66,12 +66,12 @@
   version = "v2.11.1"
 
 [[projects]]
-  digest = "1:448bf60e9dce6d5fd9dfa8fd9eb9a23163424ff4e4f142456405781e942e3998"
+  digest = "1:ed5a8df9a9f0d237764c25da704f095cbd108271c1168cb7866e3e78577110c4"
   name = "github.com/evanphx/json-patch"
   packages = ["."]
   pruneopts = "NT"
-  revision = "026c730a0dcc5d11f93f1cf1cc65b01247ea7b6f"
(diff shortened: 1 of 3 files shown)


aslakknutsen:
Could have a test... but .. When does this issue come up?

bartoszmajsak:
> Could have a test... but .. When does this issue come up?

Coming from a transitive update of, presumably, jsonpatch and then it's failing for missing `replica` path. So it is indirectly covered for the negative case. We don't have a deployment in test having that path at all - we assume it's there.

aslakknutsen:
I meant where do we see it fail. Why would replicas be missing?

bartoszmajsak:
> I meant where do we see it fail. Why would replicas be missing?

In unit tests - e.g. `k8s/deployment_test.go`

aslakknutsen:
ah ic what you mean.

PR #344, opened 2020-01-10
feat: logs version when starting the operator

bartoszmajsak:

Logs version of the operator in the container logs.

I'm mostly using `k9s` to fiddle around, so I thought it will make my life a bit easier - I know instantly which version I'm running against (especially when using `latest` as image tag)

- logs version when starting operator
- renames `PrintVersion` to `LogVersion`


bartoszmajsak:
I noticed the following error in the operator logs while making this little improvement:

```
{"level":"info","ts":1578653676.316736,"logger":"serve","msg":"failed to create or get service for metrics: services \"istio-workspace-metrics\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}     
```

is that something we should take a closer look at?

bartoszmajsak:
Issue opened https://github.com/Maistra/istio-workspace/issues/345

PR #391, opened 2020-03-19
feat(cmd): version cmd prints short semver info [skip-e2e]

### LINKED ISSUE #390: Print just version of the ike binary
##### What can be improved

At the moment we print lots of details about the binary build when asking for `version`:

```
╰─ ike version
Ike Version: v0.0.2-next-25433a6-dirty
Go Version: go1.14
Go OS/Arch: linux/amd64
operator-sdk Version: v0.15.2
Build Commit: 25433a6-dirty
Build Time: 2020-03-19T08:42:19Z
```

We should have a way to simply print `v0.0.2-next-25433a6-dirty`

```
╰─ ike version -s|--short
v0.0.2-next-25433a6-dirty
```

##### Why

It could be handy for scripting purposes.

bartoszmajsak:

Provides `-s|--short` flag for `ike version`

Closes #390

### DIFF
--- a/pkg/cmd/version/version.go
+++ b/pkg/cmd/version/version.go
@@ -15,16 +15,31 @@ var log = logf.Log.WithName("version")
 
 // NewCmd creates version cmd which prints version and Build details of the executed binary
 func NewCmd() *cobra.Command {
-	return &cobra.Command{
+	versionCmd := &cobra.Command{
 		Use:          "version",
 		Short:        "Prints the version number of ike cli",
 		Long:         "All software has versions. This is Ike's",
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			LogVersion()
+			short, err := cmd.Flags().GetBool("short")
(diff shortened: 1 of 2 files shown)


pkg/cmd/version/version.go
```diff
@@ -15,16 +15,31 @@ var log = logf.Log.WithName("version")
...
+			}
 			return nil
 		},
 	}
+
+	versionCmd.Flags().BoolP("short", "s", false, "prints only version without build details")
+	return versionCmd
+}
+
+func logShortVersion() {
```
aslakknutsen: Why is this private but the other public?

bartoszmajsak: b/c the other is called from the operator side (`ike serve`), so not only used by the command itself.

aslakknutsen: bob bob bob

bartoszmajsak: I know... but not sure anything else would be less bob bob bob

PR #400, opened 2020-03-24
docs: improvements identified during the release

bartoszmajsak:

- not tagging as part of `scripts/release.sh`
- updated docs about the release process
- minor fixes observed during the release of `v0.0.3`


docs/modules/ROOT/pages/dev_guide.adoc
```diff
@@ -1,22 +1,50 @@
...
+This will validate if above steps were correctly performed and create two commits:
+
+. `release: v0.0.2`
+  * in this commit autogenerated changelog will be appended to release notes
+  * the version in `docs/antora.yml` will be changed to a matching version
+  * release notes will be prepended to `docs/modules/ROOT/pages/release_notes.adoc`
+. `release: next iteration`
+  * reverts version-related changes from previous commit
+
+IMPORTANT: Not using `--dry-run` flag will result in pushing release commits.
```
aslakknutsen: Just a question; Since running this is part of the manual process, and it comes with a warning.. Couldn't we just switch it around: If you run it without a --do-release it does a dry run. That way it's a more conscious action to make the release (it will be scripted anyway so 'we' won't use the do-release flag.)

bartoszmajsak: Agreed, lemme add that part. Generally speaking, I want that to be fully automated soon, but until then... :)

bartoszmajsak: e4874be

PR #430, opened 2020-05-06
chore(e2e): removes openshift 3.11 tests infra

aslakknutsen:


e2e/e2e_suite_test.go
```diff
@@ -43,18 +42,6 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	tmpPath.SetPath(path.Dir(shell.CurrentDir())+"/dist", os.Getenv("PATH"))
 	ensureRequiredBinaries()
 
-	shouldManageCluster := manageCluster()
```
bartoszmajsak, line 46: In addition `ClientVersion` as-is is not really useful for anything with this change - it will always be `4`. See line `40` here.

https://github.com/Maistra/istio-workspace/blob/1a342b0a7f9dae4c9ceee19037d14c9091a13d25/e2e/infra/cluster_operations.go#L45-L70

bartoszmajsak: Implies cleanup in docker registry logic too.

e2e/e2e_suite_test.go
```diff
@@ -43,18 +42,6 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
-		executeWithTimer(func() {
-			fmt.Printf("\nStarting up Openshift/Istio cluster in [%s]\n", tmpClusterDir)
-			projectDir := testshell.GetProjectDir()
-			Expect(os.Setenv("IKE_CLUSTER_DIR", tmpClusterDir)).ToNot(HaveOccurred())
-			<-testshell.ExecuteInDir(projectDir, "make", "start-cluster").Done()
-		})
-	}
-
 	executeWithTimer(func() {
 		LoginAsTestPowerUser()
```
bartoszmajsak, line 47: Also this needs cleanup:

https://github.com/Maistra/istio-workspace/blob/1a342b0a7f9dae4c9ceee19037d14c9091a13d25/e2e/e2e_suite_test.go#L62-L66

e2e/e2e_suite_test.go
bartoszmajsak: 90908bf

e2e/e2e_suite_test.go
bartoszmajsak: 90908bf

PR #449, opened 2020-05-22
feat(cli): if no ports are defined, rely on auto expose from tp

### LINKED ISSUE #35: Auto expose ports based on D/DC port setup
The D/DC config found on the server would have some ports defined. The `ike` CLI would reuse this information to auto set the `--port` parameter.

If you have multiple ports exposed - exposed all of them - get free local ports to be mapped to D/DC.

List all the exposed ports as part of the command output.

aslakknutsen:
fixes #35

### DIFF
--- a/pkg/cmd/develop/develop.go
+++ b/pkg/cmd/develop/develop.go
@@ -77,7 +77,7 @@ func NewCmd() *cobra.Command {
 	}
 
 	developCmd.Flags().StringP("deployment", "d", "", "name of the deployment or deployment config")
-	developCmd.Flags().StringP("port", "p", "8000", "port to be exposed in format local[:remote]")
+	developCmd.Flags().StringP("port", "p", "", "port to be exposed in format local[:remote]")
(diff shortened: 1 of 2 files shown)


aslakknutsen:
@bartoszmajsak Not sure if we want to default to 8080 if tp pre 0.106 or something?

pkg/cmd/develop/develop.go
```diff
@@ -125,11 +125,17 @@ func parseArguments(cmd *cobra.Command) []string {
...
 	}
 
-	tpCmd := append([]string{
+	tpArgs := []string{
 		"--deployment", cmd.Flag("deployment").Value.String(),
-		"--expose", cmd.Flag("port").Value.String(),
 		"--method", cmd.Flag("method").Value.String(),
-		"--run"}, runArgs...)
+	}
+	if s, err := cmd.Flags().GetString("port"); err == nil && s != "" {
```
bartoszmajsak, line 132: I think that should work too:

```suggestion
	if cmd.Flags().Changed("port") {
```

PR #453, opened 2020-05-25
feat(tests): builds and pushes test images before the test suite starts

### LINKED ISSUE #313: Optimize test images usage
##### What can be improved

We should build and push images used in tests before the test suite starts.

##### Why

At this moment, each test builds and pushes images using its own namespace as a repository. That makes tests isolated, but also unnecessary slow. We should build all images upfront and push them to the cluster repository as a pre-test step in the pipeline.

bartoszmajsak:
### Short description of what this resolves:

Builds and pushes test images before the test suite starts

- cleaned up build / push image logic
- removed unnecessary logins
- minor clean up of infra files

Fixes #313

### DIFF
--- a/e2e/e2e_suite_test.go
+++ b/e2e/e2e_suite_test.go
@@ -24,6 +24,9 @@ func TestE2e(t *testing.T) {
 	RunSpecWithJUnitReporter(t, "End To End Test Suite")
 }
 
+const PreparedImageV1 = "prepared-image"
+const PreparedImageV2 = "image-prepared"
+
 var tmpClusterDir string
 
 var tmpPath = NewTmpPath()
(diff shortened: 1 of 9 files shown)


e2e/e2e_suite_test.go
```diff
@@ -46,7 +49,13 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 		fmt.Printf("\nExposing Docker Registry\n")
 		<-testshell.Execute(`oc patch configs.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge`).Done()
 
+		<-testshell.Execute(NewProjectCmd(ImageRepo)).Done()
+		UpdateSecurityConstraintsFor(ImageRepo)
+
 		BuildOperator()
+		BuildTestService()
```
aslakknutsen, line 56: Could we make these lazy instead? Would make it more convenient when running a single test to no have to build 'everything'.

bartoszmajsak: I can see how that could be done nicely - so far only ugly solutions are coming to mind.

Just wondering how frequently that would be a pain. Running a single test from the suite anyway will require to put `XDescribe` on all the other files but the one you do `FIt`, right? So in such a case you can just comment that one out too?

aslakknutsen: You only need to do Fxx on the one you want, no need to Xxx anything. You can Focus from cli as well, but normally easier to just do it in code. 

I did the 'comment out' strategy on the auto complete stuff in global scope, so yea, doable, but ..

aslakknutsen: As far as ugly, we only need a few global bools'.

PR #483, opened 2020-06-24
feat(operator): expose host route in session status

### LINKED ISSUE #242: Expose external host name in Session.status

aslakknutsen:
Allow a client to know which session routes have been exposed for this
session.

fixes #242

### DIFF
--- a/pkg/istio/gateway_test.go
+++ b/pkg/istio/gateway_test.go
@@ -91,6 +91,17 @@ var _ = Describe("Operations for istio gateway kind", func() {
 				Expect(gw.Spec.Servers[0].Hosts).To(ContainElements("domain.com", "test.domain.com"))
 			})
 
+			It("add single session - verify ref", func() {
+				err := istio.GatewayMutator(ctx, ref)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(ref.ResourceStatuses).To(HaveLen(1))
+				Expect(ref.ResourceStatuses[0].Name).To(Equal("gateway"))
+				Expect(ref.ResourceStatuses[0].Kind).To(Equal("Gateway"))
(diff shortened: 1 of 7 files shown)


pkg/istio/gateway_test.go
```diff
@@ -91,6 +91,17 @@ var _ = Describe("Operations for istio gateway kind", func() {
...
 
+			It("add single session - verify ref", func() {
+				err := istio.GatewayMutator(ctx, ref)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(ref.ResourceStatuses).To(HaveLen(1))
+				Expect(ref.ResourceStatuses[0].Name).To(Equal("gateway"))
+				Expect(ref.ResourceStatuses[0].Kind).To(Equal("Gateway"))
+				Expect(ref.ResourceStatuses[0].Prop).ToNot(BeEmpty())
+				Expect(ref.ResourceStatuses[0].Prop["hosts"]).To(Equal("test.domain.com"))
```
bartoszmajsak, line 102: Isn't `Expect(ref.ResourceStatuses[0].Prop["hosts"]).To(Equal("test.domain.com"))` enough? 101 seems somewhat redundant.

aslakknutsen: technically...

But the error will be "map is empty" as opposed to " '' doesn't match "

aslakknutsen: then again if more variables are added it would fail the Empty() check..

aslakknutsen: https://github.com/maistra/istio-workspace/pull/483/commits/35f1d9e9b5c6ec5ce165e824f0f43eabfab25b2c

PR #489, opened 2020-07-02
fix(cli): return structural response on cli

aslakknutsen:
`ike create -i x::z -d reviews-v1 --json`
```json
{
  "name": "reviews-v1",
  "strategy": "prepared-image",
  "args": {
    "image": "x::z"
  },
  "targets": [
    {
      "kind": "Deployment",
      "name": "reviews-v1",
      "action": "located",
      "labels": {
        "app": "reviews",
        "version": "v1"
      }
    },
    {
      "kind": "Service",
      "name": "reviews",
      "action": "located",
      "labels": {
        "app": "reviews"
      }
    },
    {
      "kind": "Gateway",
      "name": "test-gateway",
      "action": "located",
      "labels": {
        "ike.hosts": "aslak.test.io"
      }
    }
  ],
  "resources": [
    {
      "kind": "Deployment",
      "name": "reviews-v1-v1-aslak-iphps",
      "action": "created"
    },
    {
      "kind": "DestinationRule",
      "name": "reviews",
      "action": "modified"
    },
    {
      "kind": "Gateway",
      "name": "test-gateway",
      "action": "modified",
      "prop": {
        "hosts": "aslak-iphps.aslak.test.io"
      }
    },
    {
      "kind": "VirtualService",
      "name": "reviews",
      "action": "modified"
    },
    {
      "kind": "VirtualService",
      "name": "productpage-aslak-iphps",
      "action": "created"
    }
  ]
}
```

`ike version`
```
Ike Version: v0.0.3-next-e1cc2c2
Go Version: go1.13
Go OS/Arch: linux/amd64
operator-sdk Version: v0.17.1
Build Commit: e1cc2c2
Build Time: 2020-07-06T12:46:36Z
```
Related to #240

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/maistra/istio-workspace/pkg/cmd/completion"
 	"github.com/maistra/istio-workspace/pkg/cmd/config"
 	"github.com/maistra/istio-workspace/pkg/cmd/format"
@@ -17,7 +18,9 @@ import (
 	"github.com/spf13/viper"
 )
 
-var logger = log.Log.WithValues("type", "root")
+var logger = func() logr.Logger {
+	return log.Log.WithValues("type", "root")
+}
 
 // NewCmd creates instance of root "ike" Cobra Command with flags and execution logic defined.
 func NewCmd() *cobra.Command {
@@ -30,6 +33,10 @@ func NewCmd() *cobra.Command {
 			"For detailed documentation please visit https://istio-workspace-docs.netlify.com/\n\n",
 		BashCompletionFunction: completion.BashCompletionFunc,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if json, _ := cmd.Flags().GetBool("json"); json {
+				log.SetLogger(log.NilLog)
+				cmd.SilenceErrors = true
+			}
 			if v.Released() {
 				go func() {
 					latestRelease, _ := version.LatestRelease()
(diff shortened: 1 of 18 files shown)


aslakknutsen:
@bartoszmajsak I guess one question here would be; Should the son dump be of the complete Session object or only of the status.ref that the command is referring to... I'm leaning towards the latter.

bartoszmajsak:
> @bartoszmajsak I guess one question here would be; Should the json dump be of the complete Session object or only of the status.ref that the command is referring to... I'm leaning towards the latter.

I think we should output necessary things and not more right now - we can always extend it later if there will be a need.

bartoszmajsak:
Can you paste sample output in the PR description?

Namely:

- new json output
- ike version (just want to see to check/ensure)

pkg/cmd/root.go
```diff
@@ -30,6 +34,13 @@ func NewCmd() *cobra.Command {
 			"For detailed documentation please visit https://istio-workspace-docs.netlify.com/\n\n",
 		BashCompletionFunction: completion.BashCompletionFunc,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			cmd.Flags().VisitAll(func(flag *pflag.Flag) {
+				if flag.Changed && strings.Join(flag.Annotations["silent"], "") == "true" {
```
bartoszmajsak, line 38: Can `flag.Annotations["silent"]` panic?

pkg/log/log.go
```diff
@@ -15,8 +15,11 @@ import (
 	zapr "sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
+// NilLog is an non logger logger.
```
bartoszmajsak, line 18:
```suggestion
// NilLog is a no-op logger.
```

pkg/cmd/root.go
aslakknutsen: No, it should be ok. You can read from an uninitialized map, but not write to one.

PR #492, opened 2020-07-07
feat(cli): revert ref back to original on join

### LINKED ISSUE #289: Ability to use ike develop on a PR preview
##### What can be improved

When a CI job has created a session with a prepared-image reference, a user can choose to take over the session reference with a `develop` operation to e.g. debug something specific locally.

`ike` will, in this case, update the session ref to a telepresence session on join, but on leave delete the session as it's the only participant. `ike develop` should probably revert back to the prepared-image so the 'preview' still works when the developer is done debugging.

##### Why

Not reverting back to the prepared-image, even tho next commit to the pull request would recreate it, will leave the 'preview' in same state as 'production'. That's because the session will be gone after finishing `ike develop`, but PR will still refer to prepared-image.

aslakknutsen:
When e.g. prepared-image is used by a PR session, a developer can join the
same session and overtake the same ref with a telepresence strategy. On develop
exit, it will now revert the ref back to its original state.

Fixes #289

- [x] More test coverage of session Handlers


pkg/cmd/internal/session/session_func.go
```diff
@@ -78,7 +78,9 @@ func ToOptions(flags *pflag.FlagSet) (session.Options, error) {
 		}
 	}
 
+	revert := commandName == "develop"
```
bartoszmajsak, line 81: Just wondering if we can't have an annotation on the cmd itself defining it's revertable?

pkg/cmd/internal/session/session_func.go
```diff
@@ -18,7 +18,7 @@ func Sessions(cmd *cobra.Command) (session.State, session.Options, func(), error
 		sessionHandler = session.CreateOrJoinHandler
 	}
 
-	options, err := ToOptions(cmd.Flags())
+	options, err := ToOptions(cmd.Name(), cmd.Flags())
```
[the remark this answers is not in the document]
bartoszmajsak: But if there is no elegant way to rework `removeFlagFromSet` used in related tests I'm fine with keeping it like it is.

pkg/internal/session/session_test.go
```diff
@@ -40,3 +217,50 @@ var _ = Describe("Session operations", func() {
...
+
+// helper function to mimic the server side reacting to the session object.
+func addSessionRefStatus(c *session.Client, sessionName string) func() {
+	done := false
+	go func() {
+		for {
+			if done {
+				break
+			}
+			time.Sleep(100 * time.Millisecond)
```
bartoszmajsak, line 229: Why `100`? Is it significant? Also - why sleeping here?

pkg/internal/session/session.go
```diff
@@ -29,6 +29,8 @@ type Options struct {
 	RouteExp       string            // expression of how to route the traffic to the target resource
 	Strategy       string            // name of the strategy to use for the target resource
 	StrategyArgs   map[string]string // additional arguments for the strategy
+	Revert         bool              // Revert back to previous known value if join/leave a existing session with a known ref
+	Client         *Client           // Alternative client to be used. If nil a DefaultClient is created
```
bartoszmajsak, line 33: I am not sure `Client` should belong to this data structure, to be honest. Can we simply pass it to `Handler` when creating it?

pkg/cmd/internal/session/session_func.go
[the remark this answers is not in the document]
aslakknutsen: Yeah, thought about passing cmd directly, but there is some value in passing the leat amount of info down as opposed to the whole world. The command name check there doesn't really make much sense in the greater context, but not sure what else to use.

pkg/cmd/internal/session/session_func.go
aslakknutsen: Ah yea, that's not a bad idea.

pkg/internal/session/session_test.go
aslakknutsen: it's a while loop. It needs to take a break to not kill the cpu. :) 100 is semi random, but cli waits 1s pr check.

pkg/internal/session/session.go
aslakknutsen: both client and handler were private structs and are not normally known outside. I found it more sensible to be able to swap the Client impl then the Handler impl.

bartoszmajsak: My point was more about where the `Client` instance should belong. `Options` feels to me like plain data structure and holding an instance of the `Client` makes a bit more than it should be in my eyes. Thus my thinking was that `Client` could be passed when we call `CreateOrJoin` or any other func which creates instance of the `Handler`.

pkg/internal/session/session_test.go
bartoszmajsak: But what are we waiting for here? My thinking is that every test relying on it will add `100ms` penalty which might not be necessary. Or am I missing something here?

pkg/cmd/internal/session/session_func.go
bartoszmajsak: go-for-it

pkg/internal/session/session.go
aslakknutsen: Ah, sorry, I mixed Handler and handler :)

We can make it `type Handler func(opts Options, client Client) (State, func(), error)` for sure.

pkg/cmd/internal/session/session_func.go
aslakknutsen: https://github.com/maistra/istio-workspace/pull/492/commits/620b3b6b52f88b6807d3b6ca84f66822fb32df85

pkg/internal/session/session.go
aslakknutsen: https://github.com/maistra/istio-workspace/pull/492/commits/9b531bd991705e7e12689f5fd4f45c966730c78a

pkg/internal/session/session_test.go
aslakknutsen: https://github.com/maistra/istio-workspace/blob/master/pkg/internal/session/session.go#L178

* The Client creates a Session object
* The Client starts a wait cycle of 1s*10 looking for the Status field 
* The Operator reacts to the Session object and creates the Status field (the function simulates this)
* The Client notices the Status field and returns

I couldn't find an easy way to make a 'callback' on Create/Update or similar, so both sides 'pull'. The 'wait' is that the Session object might not be created yet when this starts up, or it will at some point change. So it checks every 100ms to see if there is something to do.

aslakknutsen: It doesn't matter if that is 100ms or 1ms, both are to slow to take part in the first Client check, which means anything <1s is fine.  The combined added time for these are 4s.

aslakknutsen: We could add a time.Duration to our Options to lower the 1s wait time for tests tho..

aslakknutsen: https://github.com/maistra/istio-workspace/pull/492/commits/72892c8a381cce63c781f8d2a443419c07b2255d

That makes it 0.04s instead of 4s

PR #515, opened 2020-08-19
fix: temp directory creation in get-ike for OSX

robertpanzer:
The get-ike script, when ran without the `--dir` option, doesn't seem to run on my OSX Mojave, and probably also not on Catalina, as it calls `mktemp --suffix`.
Instead of that `mktemp` on OSX takes an option `-t` that defines a prefix for the temporary name:

```
# curl -sL http://git.io/get-ike | bash
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
```

This PR tries to fix this by calling `mktemp -t` instead of `mktemp --suffix` on Mac.

Fixes #

### DIFF
--- a/scripts/get.sh
+++ b/scripts/get.sh
@@ -102,7 +102,13 @@ test -z "$version" && version="$(last_version)"
 test -z "$version" && {
   die "Unable to get ike version. You can still try to download it manually from $RELEASES_URL."
 }
-test -z "$dir" &&  dir="$(mktemp -d --suffix=-ike-${version})"
+test -z "$dir" && {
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
Thanks @robertpanzer. The reason why e2e tests on Openshift are failing is that it runs the build against forked repo and thus is not sharing env variables we have for connecting to the cluster. We are looking at the solution for that.

bartoszmajsak (approved):
VD4kHK0GUbmoOcWYyQ

PR #516, opened 2020-08-20
Operator failure during multiple ref addition to session

aslakknutsen:

* Fixes an error in the log about VirtualService creation
* Returns added Gateway host even if already added

* Verify if VirtualService failed or not
* Always add host even if it wasn't really added during this reconcile


pkg/controller/session/session_controller_int_test.go
```diff
@@ -234,6 +234,51 @@ var _ = Describe("Complete session manipulation", func() {
 				gw = get.Gateway("test", "test-gateway")
 				Expect(gw.Spec.Servers[0].Hosts).To(HaveLen(2))
 			})
+
+			It("mutation should be successful", func() {
```
bartoszmajsak, line 238:
```suggestion
			It("should mutate all refs added to Session", func() {
```

pkg/istio/gateway_test.go
```diff
@@ -143,6 +170,21 @@ var _ = Describe("Operations for istio gateway kind", func() {
 				Expect(gw.Spec.Servers[0].Hosts).To(HaveLen(2))
 				Expect(gw.Spec.Servers[0].Hosts).To(ContainElements("domain.com", "test.domain.com"))
 			})
+
+			It("only return added hosts once", func() {
```
bartoszmajsak, line 174:
```suggestion
			It("should only return added hosts once", func() {
```

bartoszmajsak (approved):
Minor suggestions - nitpicking :)

PR #529, opened 2020-09-18
feat(operator): add 'existing' strategy

### LINKED ISSUE #527: Avoid cloning a target deployment when deployment exist
#### Why

In the Che Workspace routing case we're targeting an existing Deployment so no cloning is required.

#### What

Suggestion to add a new ref strategy "existing" that does not invoke the clone operation in the Deployment/DeploymentConfig mutators

aslakknutsen:
A strategy to bypass the 'clone' operation. Intended to be used to route
traffic directly to an existing deployment.

Fixes #527

### DIFF
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -8,6 +8,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	// StrategyExisting holds the name of the existing strategy
+	StrategyExisting = "existing"
+)
+
 // SessionContext holds the context for a single session object, giving access to key things like REST Client and target Namespace.
 type SessionContext struct {
 	context.Context
--- a/pkg/k8s/deployment.go
+++ b/pkg/k8s/deployment.go
@@ -56,6 +56,10 @@ func DeploymentMutator(ctx model.SessionContext, ref *model.Ref) error {
 	}
 	ctx.Log.Info("Found Deployment", "name", deployment.Name)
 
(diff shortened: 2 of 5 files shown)


pkg/model/model.go
```diff
@@ -8,6 +8,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	// StrategyExisting holds the name of the existing strategy
+	StrategyExisting = "existing"
```
bartoszmajsak, line 13: What about the other strategies? Do we want to add them too?

bartoszmajsak (commented):
Something to add to the docs too?

pkg/model/model.go
aslakknutsen: They are supposed to be dynamic. The only thing that would reference them would be test code. The thinking is to read in 'most' strategies from a ConfigMap to configure the Template engine. Existing is a little bit of a hack as it affects other parts of the code than the other strategies. I couldn't really find a nice way to make them all the same unless we want to introduce a second handler (but then there is a Deployment vs DeploymentConfig issue as well) of some sort like telepresence+clone.

aslakknutsen:
> Something to add to the docs too?

I would consider this strategy a bit of an internal one for now since it requires something pretty special to do part of the work that is outside of the normal operator flow.

bartoszmajsak (approved):
Once cluster is back and healthy let's re-run tests and merge it.

Issue #528, opened 2020-09-16
Should be able to manipulate the creation of a Deployment


aslakknutsen:
#### Why

Che Server will launch a Workspace by creating a Deployment object. For Ike to route the traffic from a specific Service to this new Deployment we need the ability to manipulate the Deployment object.

#### What

A MutatingAdmission WebHook should give us the option to add the target Deployment labels to the Workspace Deployment and by that route traffic to it.

Only mutate Deployments that are labeled` ike.target=x`

PR #531, opened 2020-09-24
feat(client): make client wait condition configurable

aslakknutsen:
In a normal cli setting, we're looking for a Deployment or a DeploymentConfig,
but with the che integration, the Deployment already exists so we need to be
able to wait for something else.

related to #528

### DIFF
--- a/pkg/internal/session/session.go
+++ b/pkg/internal/session/session.go
@@ -23,14 +23,30 @@ var (
 
 // Options holds the variables used by the Session Handler.
 type Options struct {
-	NamespaceName  string            // name of the namespace for target resource
-	DeploymentName string            // name of the initial resource to target
-	SessionName    string            // name of the session create or join if exist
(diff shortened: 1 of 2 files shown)


pkg/internal/session/session.go
```diff
@@ -23,14 +23,30 @@ var (
...
+	}
+	return o.WaitCondition(res)
+}
+
+func defaultWaitCondition(res *istiov1alpha1.RefResource) bool {
+	if *res.Kind == "Deployment" || *res.Kind == "DeploymentConfig" {
+		return true
+	}
+	return false
 }
```
bartoszmajsak, line 50:
```suggestion
func defaultWaitCondition(res *istiov1alpha1.RefResource) bool {
	return *res.Kind == "Deployment" || *res.Kind == "DeploymentConfig" 
}
```

bartoszmajsak (approved):
Minor thing :)

PR #557, opened 2020-11-17
feat(build): introduces go modules

### LINKED ISSUE #16: Switch from dep to go modules

bartoszmajsak:

As go modules became a standard to manage project dependencies and with `dep` being not actively maintained anymore this PR switches our codebase (and toolchain) to use go modules instead.

- removes dep artifacts (Gopkg.{toml,lock}) and introduces `go.mod`
- adjusts build process (`Makefile` and CI) to use go modules 
- defines tools dependencies in `tools.go` file (with one exception of `golangci-lint` see https://github.com/golangci/golangci-lint/issues/1087)

Fixes #16


bartoszmajsak:
Ugh it seems GH changed the way how Actions work. https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

I will have to adjust it in separate PR. Coming soon!

bartoszmajsak:
I think there's one remaining bit here which needs clean - removal of `GOPATH` as we still refer to it in e.g. `Makefile`.

aslakknutsen (approved):

bartoszmajsak:
`$GOPATH` needs to stay as shell scripts used by codegen are still relying on it. When test on daily cluster passes, we can merge.

PR #612, opened 2021-01-18
fix(e2e): test-e2e does not need to depend on compile step

aslakknutsen:


bartoszmajsak (approved):

aslakknutsen:
Test suite depends on ike being on path on suite level. Technically docker-build called from test suite will make it, but it's a bit magical.

PR #651, opened 2021-02-03
fix(log): add log caller to structured logger

aslakknutsen:


bartoszmajsak (approved):
Zap! 

3ohs4wYPap57wUiJcQ

PR #667, opened 2021-02-05
release: add bundle push step to release process

aslakknutsen:
/skip-e2e

`make bundle` will update `config/manager/kustomization.yaml` with the new v. In theory, this should be committed, but is it really needed as it will be regenerated on each run anyway?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -230,7 +230,7 @@ jobs:
       - run:
           name: "Push it!"
           command: |
-            make deps tools docker-build docker-push-versioned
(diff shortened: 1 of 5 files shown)


alien-ike:
Skipping CI for Draft Pull Request.
If you want CI signal for your change, please convert it to an actual PR.
You can still manually trigger a test run with `/test all`

bartoszmajsak:
For consistency, we should make this part of the tagged commit IMHO

bartoszmajsak (approved):

PR #675, opened 2021-02-08
fix(bundle): sets IKE_VERSION for making bundle

bartoszmajsak:

`IKE_VERSION` sets explicitly, as tag does not exist yet on the commit for which bundle should be made.

- `release.sh` updates


bartoszmajsak:
Closing as we convinced GH to 500 on that given PR. New one recreated as #676

Issue #491, opened 2020-07-03
Deploy operator to Catalog


bartoszmajsak:
##### What can be improved

We have OLM generation, now we should deploy it to the Catalog.

##### Why

To follow official guidelines of installation in OpenShift and ultimately get in to OperatorHub

bartoszmajsak:
- `ike install` should stay as we are not only targeting OpenShift with catalog
- we should find a way to automatically test the correctness of the installation
- (but also) cover both paths

PR #689, opened 2021-02-11
fix(release): enables operatorhub signed commits

bartoszmajsak:

We were trying to sign git commit used to raise PRs for operator-hub incorrectly. This PR fixes that and adds few other improvements to the process.

- `git commit -S` is using gpg sign with passphrase set in non-interactive mode
- prefixes branch name with target hub name to avoid conflict of pushing to the same branch for different PRs
- removes configmap from bundle
- `make clean` wipes out `bundle/` too
- enables `GORELEASER_OPTS` env var for special params - this is especially useful while testing on another fork (we can skip validation or publishing for example)
- deletes tmp git clone folder at the end of the script 
- `operatorhub.sh` can now define test suite to be executed by using `-t kiwi` or `--test=orange` for example

Related to #491


bartoszmajsak:
Example of raised PRs https://github.com/maistra/community-operators/pulls

aslakknutsen (approved):

PR #763, opened 2021-03-24
feat(circleci): caches snap deps

bartoszmajsak:

Snap has been quite unstable for couple of weeks. This PR brings caching of snaps so they are downloaded once and then shared across the builds.

- downloads snaps locally to install later from files
- caches snaps for each job separately

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,7 +99,7 @@ jobs:
       - run:
           name: "Obtain snap cache version"
           command: |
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
#### Caveat 1

I am not sure if that mitigates the issue with snap being offline, as there are checks performed during the install from files.

#### Caveat 2

Versioning by channel (eg. `1.16/stable` for golang) is a moving target but caching will lock it on the certain version - when job has been executed for the first time and cached. I can think of using revisions in calculating hash/version as they are by default added to the downloaded snap file e.g. `golang_1234234.snap`.

aslakknutsen (approved):

bartoszmajsak:
Ad. ***Caveat 1*** - once you have `.snap` and `.assert` downloaded you can install while being totally offline. 

As for versioning (***Caveat 2***) - lemme see what I can do with that.

PR #792, opened 2021-04-07
fix(lint): enables goerr113

### LINKED ISSUE #480: Fix goerr113 errors
##### What can be improved

`goerr113` has been temporarily disabled with changes brought by #476. Enable it and fix errors.

##### Why

Starting from Go 1.13 the standard `error` type behaviour was changed: one `error` could be derived from another with `fmt.Errorf()` method using `%w` format specifier.

So the errors hierarchy could be built for flexible and responsible errors processing.

And to make this possible at least two simple rules should be followed:

* `error` values should not be compared directly but with `errors.Is()` method.
* `error` should not be created dynamically from scratch but by the wrapping the static (package-level) error.

This linter is checking the code for these 2 rules compliance.

`err113` reports every `==` and `!=` comparison for exact `error` type variables except comparison to `nil` and `io.EOF`.

Also, any call of `errors.New()` and `fmt.Errorf()` methods are reported except the calls used to initialise package-level variables and the `fmt.Errorf()` calls wrapping the other errors.

bartoszmajsak:
Fixes #480

### DIFF
--- a/pkg/reference/enqueue_annotations.go
+++ b/pkg/reference/enqueue_annotations.go
@@ -131,10 +131,10 @@ func (e *EnqueueRequestForAnnotation) Generic(evt event.GenericEvent, q workqueu
 // Annotations are accumulated as a comma-separated list.
 func Add(owner types.NamespacedName, object client.Object) error {
 	if owner.Namespace == "" {
-		return fmt.Errorf("%T does not have a namespace, cannot call Add", owner)
+		return fmt.Errorf("%T does not have a namespace, cannot call Add", owner) //nolint:goerr113 //reason useful to have owner in error
 	}
 	if owner.Name == "" {
-		return fmt.Errorf("%T does not have a name, cannot call Add", owner)
+		return fmt.Errorf("%T does not have a name, cannot call Add", owner) //nolint:goerr113 //reason useful to have owner in error
 	}
 
 	annotations := object.GetAnnotations()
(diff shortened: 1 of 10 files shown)


bartoszmajsak:
I think we can simply merge locally and amend the commit message with `/skip-e2e` when #789 is merged.

aslakknutsen (approved):

pkg/reference/enqueue_annotations.go
```diff
@@ -131,10 +131,10 @@ func (e *EnqueueRequestForAnnotation) Generic(evt event.GenericEvent, q workqueu
 // Annotations are accumulated as a comma-separated list.
 func Add(owner types.NamespacedName, object client.Object) error {
 	if owner.Namespace == "" {
-		return fmt.Errorf("%T does not have a namespace, cannot call Add", owner)
+		return fmt.Errorf("%T does not have a namespace, cannot call Add", owner) //nolint:goerr113 //reason useful to have owner in error
 	}
 	if owner.Name == "" {
-		return fmt.Errorf("%T does not have a name, cannot call Add", owner)
+		return fmt.Errorf("%T does not have a name, cannot call Add", owner) //nolint:goerr113 //reason useful to have owner in error
```
aslakknutsen, line 137: Does this check only allow static errors? A Error creating function isn't allowed in some way for dynamic data?

bartoszmajsak: Yeah, func creating error should work. Reason I didn't touch this code much is that it's "ported" from operator-lib.

Issue #827, opened 2021-05-01
istio-workspace-operator restart after launch "ike develop" command


jorgermp:
##### Issue Overview

I'm trying to use ike to make some test, but I don't work.
When I launch this command "ike develop --deployment nginx --port 80 --watch --run 'docker run -it -p 80:80 nginx' --route header:end-user=alien-ike --watch" I obtain this error in istio-workspacer-operator pod and restart it.
<details>
`{"level":"info","ts":1619887917.6047382,"logger":"root","caller":"session/session_controller.go:163","msg":"Reconciling Session","type":"controller","Request.Namespace":"default","Request.Name":"jorge.martinez-wsoqj"}
{"level":"info","ts":1619887917.6047118,"logger":"root","caller":"reference/enqueue_annotations.go:254","msg":"Reference Object created","type":"enqueue","kind":"/, Kind=","namespace":"default","name":"nginx-b23a6a84-jorge.martinez-wsoqj","targets":[{"Namespace":"default","Name":"jorge.martinez-wsoqj"}]}
{"level":"info","ts":1619887917.6097875,"logger":"root","caller":"session/session_controller.go:200","msg":"Added session","type":"controller","Request.Namespace":"default","Request.Name":"jorge.martinez-wsoqj"}
{"level":"info","ts":1619887917.6099486,"logger":"root","caller":"session/session_controller.go:272","msg":"Add ref","type":"controller","Request.Namespace":"default","Request.Name":"jorge.martinez-wsoqj","name":"nginx"}
{"level":"info","ts":1619887918.6608238,"logger":"root","caller":"v2/klog.go:916","msg":"Throttling request took 1.045354551s, request: GET:https://10.43.0.1:443/apis/config.gatekeeper.sh/v1alpha1?timeout=32s\n"}
[...]

PR #828, opened 2021-05-03
fix(error): handle no destination rules found error

aslakknutsen:
Ensure we return an error when no rules found, and when no subset found.

Related to #827

### DIFF
--- a/pkg/istio/destinationrule_test.go
+++ b/pkg/istio/destinationrule_test.go
@@ -164,7 +164,7 @@ var _ = Describe("Operations for istio DestinationRule kind", func() {
 
 		Context("missing rule", func() {
 
-			It("no rules found", func() {
+			It("should fail when no rules found", func() {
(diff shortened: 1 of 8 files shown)


pkg/istio/destinationrule_test.go
```diff
@@ -161,6 +161,22 @@ var _ = Describe("Operations for istio DestinationRule kind", func() {
 				Expect(dr.Items[0].Spec.Subsets[0].TrafficPolicy.ConnectionPool.Http.MaxRetries).To(Equal(int32(100)))
 			})
 		})
+
+		Context("missing rule", func() {
+
+			It("no rules found", func() {
```
bartoszmajsak, line 167:
```suggestion
			It("should fail when no rules found", func() {
```

bartoszmajsak (approved):
steppypants-3oEjHZ0VV7sPH1CZAk

PR #837, opened 2021-05-06
feat(log): introduce emperror for errors with details

### LINKED ISSUE #797: Evaluate switching to emperror
##### What can be improved

One error interface for wrapping, multi error and detailed errors

##### Why

* multi error
* detailed error for logging
* wrapping with stack trace

https://github.com/emperror/errors

aslakknutsen:
* include details in error where possible
* only include stacktrace at the lowest level

Fixes #797


pkg/log/emperror.go
```diff
@@ -0,0 +1,60 @@
...
+			for n := i + 2; n < len(kv); n += 2 {
+				if v2, ok := kv[n].(string); ok {
+					if v == v2 {
+						kv = append(kv[:n], kv[n+2:]...)
+						n -= 2
+					}
+				}
+			}
+		}
+	}
```
bartoszmajsak, line 35: image

pkg/log/emperror_test.go
```diff
@@ -0,0 +1,39 @@
+package log_test
```
bartoszmajsak, line 1: How are we testing things here? :)

bartoszmajsak (approved):
Just two points to bring to your attention. And a gif :)

pkg/log/emperror_test.go
aslakknutsen: Yeah, wanted to bring that up. We're not testing anything. But it's a 'structure' for manual verification? Not sureif we should just dump it or what to do with it

bartoszmajsak: Ok lets keep it like that but maybe follow-up with https://github.com/maistra/istio-workspace/issues/833 should be considered as good time of evaluating it. WDYT?

aslakknutsen: That sounds good

PR #848, opened 2021-05-24
fix(make): sets mv with --force flag to avoid prompting

bartoszmajsak:


aslakknutsen (approved):

aslakknutsen:
When does this kick in? Local only?

bartoszmajsak:
> When does this kick in? Local only?

I saw it local only and won't happen during the build in the container, but it's probably good to set it this anyway.

PR #853, opened 2021-07-05
feat(circleci): aggregates junit reports for CircleCI

bartoszmajsak:

Ginkgo produces one junit xml per suite.

CircleCI can show the test summary but expects one file with all test runs.

This PR introduces simple logic around the build step which takes ginkgo results and merges them into one file, so CircleCI can consume it.

- config for circle
- aggregation using [`junit-report-merger`](https://www.npmjs.com/package/junit-report-merger) node tool \o/

##### Some insights on why I went with node

I tried few others (included go and python-based ones but they were not working). This one in the contrary works like a charm.

Another possibility I tried was to use [`gotestsum`](https://github.com/gotestyourself/gotestsum) which is pretty neat but does not really work well with ginkgo. I also tried using `go test` in combination with `gotestsum` to run ginkgo tests, but that somehow ended up with an incomplete test report. So after all those attempts jrm seems to be the only one working easily.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,12 @@ defaults:
     command: |
       nvm install 10
       nvm use 10
+  merge-junit-reports: &merge-junit-reports
+    name: "Merges ginkgo junit test results"
+    command: |
+      sudo npm install -g junit-report-merger
+      mkdir -p test-results
+      jrm ./test-results/junit-report-combined.xml $(find . -type f -name ginkgo-test-results.xml)
   docker:
     - image: &golang-img cimg/go:1.16.5-node
   machine-conf: &machine-conf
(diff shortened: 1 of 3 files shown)


bartoszmajsak:
image

.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ defaults:
       nvm install 10
       nvm use 10
   docker:
-    - image: &golang-img cimg/go:1.16.2
+    - image: &golang-img cimg/go:1.16.5-node
```
bartoszmajsak, line 29: This is needed because we use JRM (node cli tool to merge junit xmls)

Makefile
```diff
@@ -210,32 +210,32 @@ $(PROJECT_DIR)/bin/yq:
 	GOBIN=$(PROJECT_DIR)/bin go install -mod=readonly github.com/mikefarah/yq/v4
 
 $(PROJECT_DIR)/bin/protoc-gen-go:
-	$(call header,"Installing")
```
bartoszmajsak, line 213: Minor things, can be ignored ;)

aslakknutsen (approved):

PR #855, opened 2021-07-09
feat(operator): bumps operator-sdk to latest v1.8.1

bartoszmajsak:

Bumps `operator-sdk` and olm install on microk8s.

- bumps operator-sdk to latest v1.8.1 
- olm-install on microk8s has been updated too because aforementioned issue has been resolved within the container itself
- bumps kustomize to latest release v4.2.0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -255,7 +255,7 @@ jobs:
 
             make deps tools
 
-            ./bin/operator-sdk olm install --version v1.8.1
+            ./bin/operator-sdk olm install --version v0.18.2
 
             make docker-build docker-push-versioned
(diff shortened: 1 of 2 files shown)


bartoszmajsak:
At the same time, I tried to update `operator-lib` to `v0.5.0` but that results in a weird behaviour exposed by session tests. It seems that calling `Reconcile` is resulting with the object tracker losing reference to the stored session. Could be related to update of k8s libs transitively in `go.mod`.

I am not sure if it is a side effect or bug on our side, but since we are heavily reworking this anyway on `new_operator` branch I just wanted to state it here to have a trace for further investigation.

aslakknutsen (approved):

PR #865, opened 2021-07-21
feat: adds condition when the source ref is located

### LINKED ISSUE #864: Populate the Conditions with the Locator status before Modification
##### What can be improved

We could populate the Conditions the the Locator status in false state to see what will happen

##### Why

You can see what should happen and over time that it all turns true.

bartoszmajsak:

Reports successful location of defined reference (source) so that it can be seen in e.g. Openshift Console.

- adds condition when the source ref is located
- moves condition-related to logic to its own file
- minor code cleanups

Fixes #864

### DIFF
--- a/controllers/session/conditions.go
+++ b/controllers/session/conditions.go
@@ -1,9 +1,10 @@
 package session
 
 import (
+	"strconv"
+
 	istiov1alpha1 "github.com/maistra/istio-workspace/api/maistra/v1alpha1"
 	"github.com/maistra/istio-workspace/pkg/model"
-	"strconv"
 )
 
 func addConditionForLocatedRef(session *istiov1alpha1.Session, ref model.Ref, modified *model.LocatorStatus) {
--- a/controllers/session/conditions.go
+++ b/controllers/session/conditions.go
@@ -9,7 +9,7 @@ import (
 
 func createConditionForLocatedRef(ref model.Ref, located model.LocatorStatus) istiov1alpha1.Condition {
 	message := located.Kind + "/" + located.Name + " status " + ref.KindName.String() + ": "
-	reason := toReason(located.Action)
+	reason := "Scheduled"
 	typeStr := string(located.Action)
 	status := "true"
 
@@ -42,7 +42,7 @@ func createConditionForModifiedRef(ref model.Ref, modified model.ModificatorStat
 	}
 	status := strconv.FormatBool(modified.Success)
 
-	reason := toReason(modified.Action)
+	reason := "Applied"
 	typeStr := string(modified.Action)
 
 	return istiov1alpha1.Condition{
@@ -72,15 +72,3 @@ func cleanupRelatedConditionsOnRemoval(ref model.Ref, session *istiov1alpha1.Ses
 	}
 }
 
-func toReason(a model.StatusAction) string {
-	switch a {
-	case model.ActionCreate, model.ActionDelete, model.ActionLocated:
-
-		return "Scheduled"
-	case model.ActionModify, model.ActionRevert:
-
-		return "Applied"
-	}
-
-	return ""
-}
--- a/controllers/session/conditions.go
+++ b/controllers/session/conditions.go
@@ -76,10 +76,10 @@ func toReason(a model.StatusAction) string {
 	switch a {
 	case model.ActionCreate, model.ActionDelete, model.ActionLocated:
 
-		return "Handled"
+		return "Scheduled"
 	case model.ActionModify, model.ActionRevert:
 
-		return "Configured"
+		return "Applied"
 	}
 
 	return ""
--- a/controllers/session/conditions.go
+++ b/controllers/session/conditions.go
@@ -7,25 +7,26 @@ import (
 	"github.com/maistra/istio-workspace/pkg/model"
 )
 
-func addConditionForLocatedRef(session *istiov1alpha1.Session, ref model.Ref, modified *model.LocatorStatus) {
-	message := modified.Kind + "/" + modified.Name + " located " + ref.KindName.String() + ": "
-	reason := toReason(modified.Action)
-	typeStr := string(modified.Action)
-	located := "true"
-	session.AddCondition(istiov1alpha1.Condition{
+func createConditionForLocatedRef(ref model.Ref, located model.LocatorStatus) istiov1alpha1.Condition {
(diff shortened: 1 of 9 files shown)


test/cmd/test-service/main.go
```diff
@@ -29,10 +29,6 @@ const (
 	EnvServiceCall = "SERVICE_CALL"
 )
 
-var (
```
aslakknutsen, line 32: Why removed?

pkg/internal/session/session.go
```diff
@@ -216,7 +218,7 @@ func notDeleted(condition *istiov1alpha1.Condition) bool {
 }
 
 func deploymentOrDeploymentConfig(condition *istiov1alpha1.Condition) bool {
-	return condition.Source.Kind == "DeploymentConfig" || condition.Source.Kind == "Deployment"
+	return condition.Source.Kind == k8s.DeploymentKind || condition.Source.Kind == openshift.DeploymentConfigKind
```
aslakknutsen, line 221: Originally done to avoid a package import on client of the server side.. but.. it's the same binary at this point anyway

controllers/session/conditions.go
```diff
@@ -0,0 +1,84 @@
...
+			Ref:  ref.KindName.String(),
+		},
+		Message: &message,
+		Reason:  &reason,
+		Status:  &located,
+		Type:    &typeStr,
+	})
+}
+
+func addConditionForModifiedRef(session *istiov1alpha1.Session, ref model.Ref, modified *model.ModificatorStatus) {
```
aslakknutsen, line 28: Don't think we need to use a pointer on modified here

controllers/session/conditions.go
```diff
@@ -0,0 +1,84 @@
+package session
+
+import (
+	"strconv"
+
+	istiov1alpha1 "github.com/maistra/istio-workspace/api/maistra/v1alpha1"
+	"github.com/maistra/istio-workspace/pkg/model"
+)
+
+func addConditionForLocatedRef(session *istiov1alpha1.Session, ref model.Ref, modified *model.LocatorStatus) {
```
aslakknutsen, line 10: Don't think we need to use a pointer for modified here

controllers/session/conditions.go
aslakknutsen, line 28: Maybe we should evaluate to make these session.AddCondition(createConditionForModifiedRef)

aslakknutsen (changes requested):

pkg/internal/session/session.go
bartoszmajsak: Ok, didn't think about that.

controllers/session/conditions.go
bartoszmajsak: I was actually thinking about doing it differently. The current approach is a bit too procedural at this point. I will change it as it makes more sense.

bartoszmajsak: Fair enough - it's been like that originally hehe </lame_excuse>

test/cmd/test-service/main.go
bartoszmajsak: My super IDE told me it's unused :)

aslakknutsen: Hench the comment on it.. ;)

bartoszmajsak: LOL :) Speedy move - didnt even notice the comment :) Brought back in 5a02456

But on that note - can we document how and when are using "dev mode"? Can't really see any reference in the project. Maybe elaborate the comment will be sufficient and most accurate.

controllers/session/conditions.go
bartoszmajsak: Improved in 4877cf2

test/cmd/test-service/main.go
aslakknutsen: Does it matter if you don't read? :P hehe

Can't even remember atm. It's not in the Makefile. There is a flag you provide to go-bindata, but I must have been playing with it outside of Make when I was developing the UI there.

controllers/session/conditions.go
```diff
@@ -0,0 +1,86 @@
...
+
+	istiov1alpha1 "github.com/maistra/istio-workspace/api/maistra/v1alpha1"
+	"github.com/maistra/istio-workspace/pkg/model"
+)
+
+func createConditionForLocatedRef(ref model.Ref, located model.LocatorStatus) istiov1alpha1.Condition {
+	message := located.Kind + "/" + located.Name + " status " + ref.KindName.String() + ": "
+	reason := toReason(located.Action)
+	typeStr := string(located.Action)
+	status := "true"
```
aslakknutsen, line 14: Should this be false since we haven't done anything yet?

test/cmd/test-service/main.go
bartoszmajsak: I did search everywhere before removing.... So if you don't remember what's the value of keeping it around? :D

controllers/session/conditions.go
bartoszmajsak: My understanding was - it was successfully located. Isn't it what we should report here?

aslakknutsen: Yes and now. The action is "modify" and it's not done yet..

bartoszmajsak: If it's "yes and no" we have a bit of an abstraction problem :)

test/cmd/test-service/main.go
aslakknutsen: As I said, it's related to go-bindata and I know how that work. I just don'r remember how I used it back in the day. It's needed to use the debug feature of go-bindata, how ever you generate the go-bindata

bartoszmajsak: Restored in 5a024564

controllers/session/conditions.go
bartoszmajsak: Maybe `started`, `pending`, `running` and `finished` should be better than `true|false` status in such a case?

aslakknutsen: The overall status.state is Processing, Success, Failed. What we're talking about here is a condition; Is it done; bool

We could change the reason a bit to indicate something different from when it's processed. Like 'Ready to start'

bartoszmajsak: As discussed - reworded `reason` in 917ec03

aslakknutsen (approved):

PR #866, opened 2021-07-26
feat(operator): add validation chain that blocks on error

### LINKED ISSUE #856: Add Locator validation to verify the state of affairs before moving on
##### What can be improved

There should be a validation phase between locating the things that need manipulation and them being manipulated.

##### Why

This will allow early detection of errors like
* missing deployment
* missing destination rule
* missing virtual service

##### Additional

* Move destinationrule_test.go "missing rule::should fail when no rules found" test case into new Validate test case

aslakknutsen:
The Validation Chain will verify if the required objects are found after
locators has been invoked. If the chain is false further execution is
stopped.

The Chain will run twice, once before and once after locators creating a
status change to reflect current status starting with false.

Fixes #856

### DIFF
--- a/controllers/session/validation.go
+++ b/controllers/session/validation.go
@@ -48,16 +48,17 @@ func chainValidator(ctx model.SessionContext, ref model.Ref, session *istiov1alp
 
 func ResourceFound(kind string) Validator {
 	return func(store model.LocatorStatusStore) (string, error) {
+		targetType := "Find" + kind
 		if len(store(kind)) == 0 {
-			return kind + "Found", errors.New("no " + kind + " found")
+			return targetType, errors.New("no " + kind + " found")
 		}
 
-		return kind + "Found", nil
+		return targetType, nil
 	}
 }
 
 func TargetFound(store model.LocatorStatusStore) (string, error) {
-	typeName := "TargetFound"
+	typeName := "FindTarget"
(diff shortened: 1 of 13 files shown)


controllers/session/validation.go
```diff
@@ -0,0 +1,66 @@
...
+	istiov1alpha1 "github.com/maistra/istio-workspace/api/maistra/v1alpha1"
+	"github.com/maistra/istio-workspace/pkg/model"
+)
+
+// Validator returns a string of Type and a possible error.
+type Validator func(store model.LocatorStatusStore) (string, error)
+
+func chainValidator(ctx model.SessionContext, ref model.Ref, session *istiov1alpha1.Session, validators ...Validator) model.ModificatorController {
+	return func(store model.LocatorStatusStore) bool {
+		hasError := false
```
bartoszmajsak, line 17: if we would instead define `succeeded := true` we could use that for L26 and return in L45. maybe a bit more readable and coherent?

bartoszmajsak (approved):
Minor suggestion to consider. Otherwise LGTM

controllers/session/validation.go
aslakknutsen: Good point! https://github.com/maistra/istio-workspace/pull/866/commits/717a8a18052daef9f445694ed2b1b551dfcb23d0

PR #868, opened 2021-08-04
chore: do not show cmd output when getting tp version

bartoszmajsak:
Minor fix - now showing outup of `telepresence --version` to stdout

Related to #833


bartoszmajsak:
@aslakknutsen not sure if that can close #833 ;)

aslakknutsen (approved):

PR #908, opened 2021-09-16
build(golang): bumps golang builds to 1.17.1

bartoszmajsak:


aslakknutsen (approved):
Fedora 34 is still on 1.16.x but...

PR #920, opened 2021-09-22
fix(operator): add new groupname to csv config

aslakknutsen:
related to #891

PR #932, opened 2021-09-29
Operator metrics

### LINKED ISSUE #885: Add metrics numbers for Actions and API Server Operations
##### What can be improved

The exposed metrics should be more fine grained.

Capturing type of resources and their actions and api server calls.

##### Why

Gives a better overview over what the Operator is doing

aslakknutsen:

Adds visibility into the internals of the Operator.

Instrument the client.Client and the model.ModificatorStatusReporter to track function calls and actions segmented on namespace and kind

Fixes #885

- [ ] Currently includes https://github.com/maistra/istio-workspace/pull/920


controllers/session/instrumented_client.go
```diff
@@ -0,0 +1,123 @@
...
+func (i *InstrumentedClient) Update(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
+	return record(i.c.Update(ctx, obj, opts...), "update", i.getObjectKind(obj))
+}
+
+func (i *InstrumentedClient) Patch(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
+	return record(i.c.Patch(ctx, obj, patch, opts...), "patch", i.getObjectKind(obj))
+}
+
+func (i *InstrumentedClient) DeleteAllOf(ctx context.Context, obj client.Object, opts ...client.DeleteAllOfOption) error {
+	return record(i.c.DeleteAllOf(ctx, obj, opts...), "deleteallof", i.getObjectKind(obj))
```
bartoszmajsak, line 80: How about a slightly more readable key?

```suggestion
	return record(i.c.DeleteAllOf(ctx, obj, opts...), "delete_all_of", i.getObjectKind(obj))
```

controllers/session/session_controller.go
```diff
@@ -174,13 +174,15 @@ func (r ReconcileSession) WatchTypes() []client.Object {
 
 // Reconcile reads that state of the cluster for a Session object and makes changes based on the state read
 // and what is in the Session.Spec.
-func (r *ReconcileSession) Reconcile(c context.Context, request reconcile.Request) (reconcile.Result, error) { //nolint:cyclop,gocyclo //reason WIP
+func (r *ReconcileSession) Reconcile(orgCtx context.Context, request reconcile.Request) (reconcile.Result, error) { //nolint:cyclop,gocyclo //reason WIP
 	reqLogger := logger().WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Session")
 
+	c := NewInstrumentedClient(r.client)
```
bartoszmajsak, line 181: One general question - do we want to collect/expose those metrics all the time? How big is overhead here?

My line of thinking is that it could potentially be enabled through a flag instead of being always on, so the user can opt-in if they wish.

bartoszmajsak (approved):
LGTM - just two open points.

controllers/session/instrumented_client.go
aslakknutsen: Yea I was thinking of that, but then it kinda mixed in with status_x which is a sub field jatta jatta.. hehe

controllers/session/session_controller.go
aslakknutsen: No overhead to talk about.

controllers/session/instrumented_client.go
bartoszmajsak: Can't we have `status#x` instead?

Issue #886, opened 2021-09-03
Rely on PATCH of resources instead of PUT


aslakknutsen:
##### What can be improved

Currently the operator rely on PUT operations to replace the existing resources, this could be changed to PATCH.  

##### Why

Optimize network traffic and  a way to detect if we're actually changing anything. Under the assumption we're returning an empty patch set we don't need to apply it.

PR #937, opened 2021-10-07
feat(client-go): uses PATCH instead of PUT for istio resources

bartoszmajsak:

Instead of using `.Update` calls we can leverage `.Patch` for resources and their status. This PR addresses it for mutators - so changes of Istio resources.

Applying the same approach for the controller, where we reconcile `Session` CRD is slightly more challenging, as it requires merging strategy to patch e.g. empty list of conditions instead of omitting them.

- `.Patch` calls for Istio resources
- minor code cleanups

Related to #886


bartoszmajsak:
@aslakknutsen anything holding us back from merging this one?

bartoszmajsak:
> @bartoszmajsak Needs some manual verification that's all. I'm not sure the list handling actually works :) Testing it now

Does it mean we should think about some better tests?

PR #1029, opened 2022-07-13
fix(build): use go install to fetch controller-gen bin

aslakknutsen:

`make tools` does not manage to install `controller-gen` on go1.18 as `go get` for installing binaries is deprecated. 

https://go.dev/doc/go-get-install-deprecation

Switch to `go install`


bartoszmajsak (approved):
LGTM! I have overlooked it earlier when changing `go get`s in all the other places :)

You might want to merge `master` here first to unblock your tests when #1030 lands.

PR #1044, opened 2022-07-20
feat(e2e): skips completion tests if current shell is not bash

bartoszmajsak:


aslakknutsen (approved):

bartoszmajsak:
Turns out it always wraps command in `bash -c` ... So this is not needed. I'm stupid OR blind  :)

Fixed in #1055

PR #1055, opened 2022-07-29
feat(flag): enables limited-choice flags

bartoszmajsak:
* validates input
  * provides autocompletion

[asciicast](https://asciinema.org/a/kNRwHHpJE3srEofpjUFBO8wHi)


pkg/cmd/flag/limited_options_test.go
```diff
@@ -0,0 +1,70 @@
...
+		var testCmd *cobra.Command
+
+		BeforeEach(func() {
+			testCmd = newTestCmd("stout", "s", "ale", "a", "kolsch", "k")
+			testCmd.SilenceUsage = true
+			testCmd.SilenceErrors = true
+			NewCmd().AddCommand(testCmd)
+		})
+
+		It("should accept only defined flags using full names", func() {
```
bartoszmajsak, line 26:
```suggestion
		It("should accept defined value using full name", func() {
```

pkg/cmd/flag/limited_options_test.go
```diff
@@ -0,0 +1,70 @@
...
+			output, err := Run(testCmd).Passing("--style", "stout")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(output).To(ContainSubstring("selected style: 'stout'"))
+		})
+
+		It("should accept flag using abbreviation", func() {
```
bartoszmajsak, line 32:
```suggestion
		It("should accept defined value using abbreviated name", func() {
```

pkg/cmd/flag/limited_option.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// CreateOptions will return a slice of pflag.Value with name and associated abbreviation.
+// This limited set of values can be then bound to a cobra flag to limit choices for a given
+// flag. On top of that custom completion can be defined.
```
bartoszmajsak, line 12:
```suggestion
// CreateOptions will return a slice of pflag.Value with name and associated abbreviation.
// This limited set of values can then be bound to a cobra flag to limit choices for a given
// input. On top of that custom, completion can be defined.
```

aslakknutsen:
Doesn't seem to work in my bash.. trying to figure out what's going on. It seems to mostly suggest "ike" as an option for most things hehe

bartoszmajsak:
> Doesn't seem to work in my bash.. trying to figure out what's going on. It seems to mostly suggest "ike" as an option for most things hehe

Let me check on my end, as I am using `zsh` daily. Tests however run on `bash` in circleci. Not to mention all the commands are always executed using `bash -c` so in theory if tests are passing on your machine the completion should work in your regular usage. Which version are you on?

bartoszmajsak:
@aslakknutsen seems to work on my end.

```
$ bash --version
GNU bash, version 5.1.8(1)-release (x86_64-redhat-linux-gnu)

$ ike develop -d test -r run.sh -m 
inject-tcp  vpn-tcp     
```

My very old `.bashrc` contains only some `$PATH` definitions (even jboss-forge hehe oh the memories)

aslakknutsen:
It works. Just a confusion around "./ike" and "ike" way of executing it on my end. Bash only stuff works in both cases but completion that require binary execution only works when the binary is called correctly, in this case "./ike".

aslakknutsen (approved):

PR #1067, opened 2022-08-09
chore: introduces hook package to handle shutdowns in one place

bartoszmajsak:

Moves shutdown hook logic to a single place and exposes a way to register multiple handlers.

- new `hook` pkg with logic for registering shutdown hooks reacting on `io.Singal`
- small rework of `root` command 
- cmd tests adjustments

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -16,6 +16,7 @@ import (
 	"github.com/maistra/istio-workspace/pkg/cmd/config"
 	"github.com/maistra/istio-workspace/pkg/cmd/format"
 	"github.com/maistra/istio-workspace/pkg/cmd/version"
+	"github.com/maistra/istio-workspace/pkg/hook"
 	"github.com/maistra/istio-workspace/pkg/log"
 	v "github.com/maistra/istio-workspace/version"
 )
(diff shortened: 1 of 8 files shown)


pkg/cmd/root.go
```diff
@@ -66,7 +69,9 @@ func NewCmd() *cobra.Command {
 
 			return nil
 		},
-		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
+		PostRunE: func(cmd *cobra.Command, args []string) error {
+			hook.Close()
+
```
bartoszmajsak, line 74:
```suggestion
```

test/cmd.go
```diff
@@ -44,6 +48,11 @@ func executeCommandC(cmd *cobra.Command, args ...string) (c *cobra.Command, outp
 	cmd.Root().SetArgs(append(strings.Split(cmd.CommandPath(), " ")[1:], args...))
 	c, err = cmd.ExecuteC()
 
+	if err != nil {
+		// It is called as well in main.go on error. We need to close the channel to avoid leaking goroutine.
```
bartoszmajsak, line 52:
```suggestion
		// It is called as well in main.go on error. We need to close the channel to avoid leaking goroutine in tests.
```

aslakknutsen (approved):

PR #1079, opened 2022-09-02
fix: ensures version WARN always shows up

bartoszmajsak:

Ensures `WARN` about the latest release version available (if running an older one) always shows up. Right now if `cmd` had its own `PreRunE` this logic was overwritten. Moving it to `PersistentPreRunE` ensure it's always executed.

- Moves release check logic to `PersistentPreRunE` block of `root` command

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -30,6 +30,7 @@ func NewCmd() *cobra.Command {
 	var configFile string
 	releaseInfo := make(chan string, 1)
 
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
Not quite sure why it fails on `circle-ci` - smells like a flaky test, but cannot reproduce it locally. Different test order perhaps? ... Investigating

aslakknutsen (approved):

